PR #14, opened 2021-02-11
bugfix-splunk-use-initd-variable

arengifoc:
**BUGFIX**
- `splunk_use_initd` variable is boolean, but it was being compared as string (e.g. `splunk_use_initd == "true"`)
- Made minor changes on that variable to allow Splunk properly configure init script (SystemV or Systemd)


mason-splunk:
Thanks @arengifoc ! I've rolled this into https://github.com/splunk/ansible-role-for-splunk/pull/16 with a couple other minor changes

PR #21, opened 2021-02-19
Multiple bugfixes and updated WGET URLs

mason-splunk:


mason-splunk:
So, one thing that may be worth discussion here is this change:

> Updated the role default for splunk_use_initd from false (systemd) to true (init.d) as that is the default in Splunk v8.x when the -systemd-managed argument is not provided to the splunk enable boot-start command

This will have the unfortunate side effect of causing app deployment plays to fail when the value of splunk_use_initd has not been updated in group_vars/host_vars to match the actual method used on the host. It will also fail if boot-start is not enabled at all due to this change:

> Updated the restart splunk handler to use the service module instead of the shell module with the splunk binary

One possible workaround would be to ONLY use splunk_use_initd during upgrades, and to auto-discover the current boot-start method by using the `stat` module against the expected paths when boot-start is enabled. i.e.

- `/etc/init.d/splunk` would indicate that init.d is being used (for either full or UF)
- `/etc/systemd/system/SplunkForwarder.service` would indicate that systemd is being used with a UF
- `/etc/systemd/system/Splunkd.service` would indicate that systemd is being used with a full installation
- If none of the above 3 exist, then boot-start is not enabled, and we should handle restarts by calling the splunk binary directly with the `restart` argument

mason-splunk:
Additional [point of reference](https://community.splunk.com/t5/Deployment-Architecture/Systemd-start-restart-for-splunk-not-working-as-expected-CentOS/m-p/308140/highlight/true#M11701) for systemd vs init.d changes between Splunk versions

mason-splunk:
After discussing with @micahkemp-splunk we have a better solution for ensuring Ansible uses the correct handler. I will update tasks/main.yml tomorrow to do the following:

1. stat the init.d and systemd paths as described in the previous comment
2. Compare the result of the stat to the value of `splunk_use_initd`
3. If there is a discrepancy, call disable boot-start (will break this out into a separate task as it's now going to be used in multiple places), then include `configure_splunk_boot.yml` to re-configure it
4. If none of the expected file paths exist, call configure_splunk_boot.yml to configure boot-start
5. If the host is already in the expected state, do nothing related to boot-start
6. Proceed with the play

mason-splunk:
Edited PR description to reflect the final implementation

mason-splunk:
Tested successfully for the following scenarios:

1. Splunk installed but boot-start not configured - play fails
2. Splunk installed but init.d used while systemd configured - play fails
3. Splunk installed but systemd used while init.d configured - play fails
4. Run configure_splunk_boot.yml to remediate any of the above 3 conditions - successful
5. Splunk not installed at all - play successful with check_splunk.yml
5. Splunk installed with init.d, upgrade to new version using check_splunk.yml - successful
6. Splunk installed with systemd, upgrade to new version using check_splunk.yml - successful

roles/splunk/defaults/main.yml
```diff
@@ -6,9 +6,9 @@ slack_token: undefined
 splunk_home: auto_determined # This gets set by main.yml but we have to define it here or Ansible will complain that it is undefined
 splunk_package_url: auto_determined # This gets set by main.yml but we have to define it here or Ansible will complain that it is undefined
 splunk_package_path: ~/
-splunk_package_url_full: https://www.splunk.com/bin/splunk/DownloadActivityServlet?architecture=x86_64&platform=linux&version=8.0.6&product=splunk&filename=splunk-8.0.6-152fb4b2bb96-Linux-x86_64.tgz&wget=true
-#splunk_package_url_full: https://www.splunk.com/bin/splunk/DownloadActivityServlet?architecture=x86_64&platform=linux&version=8.0.4.1&product=splunk&filename=splunk-8.0.4.1-ab7a85abaa98-Linux-x86_64.tgz&wget=true
```
micahkemp-splunk, line 10: Is there a reason you're persisting this commented out line?  It seems it was used during testing, because it "changed" from `_full` to `_uf`.

roles/splunk/handlers/main.yml
```diff
@@ -15,16 +15,31 @@
...
+    regexp: '(start --no-prompt --answer-yes)$'
+    replace: '\1 --accept-license'
+  become: yes
+  when: splunk_use_initd
+
+- name: reload systemctl daemon
+  systemd:
+    daemon_reload: yes
+  become: yes
+  when: ansible_connection!="docker"
```
micahkemp-splunk, line 30: Thinking back to when I tried to use `systemctl` with docker, I wonder if this shouldn't be skipped when used under docker, but instead left up to the admin using this to ensure that their container works with `systemctl`, or alternatively that they use `initd` instead, which would be skipped if this was `when: not splunk_use_initd`.

roles/splunk/defaults/main.yml
mason-splunk: Ahh good catch.  It was supposed to be removed. I'll fix it.

mason-splunk: Updated.

roles/splunk/handlers/main.yml
mason-splunk: Removed conditional

micahkemp-splunk (approved):

PR #36, opened 2021-02-24
Fixed the Version parsing of splunk package request url

schneewe:
I had issue with the splunk version parsing from thee newest splunk eenterprise version package.
The regex for the request version was wrong.
splunk-8.1.2-545206cc9f70-Linux-x86_64.tgz -> matched to 8.1.2-545206


mason-splunk (approved):

mason-splunk:
Good catch. Thanks for the fix! Published in v1.2.4 release.

schneewe:
Thanks, and many thanks for the great project.

Issue #42, opened 2021-03-02
auditd restart does not work as expected in non-RedHat Linux distributions


arengifoc:
By running `service auditd condrestart` to restart the auditd service is causing the task to fail on distributions other than CentOS and RHEL

PR #43, opened 2021-03-02
Modified conditional handlers for restarting auditd in redhat and non redhat Linux distributions

arengifoc:
**BUGFIXES**
- Created to handlers for restarting auditd service
- One is for RedHat based distributions and the other one is for others (non RedHat based)
- Resolves https://github.com/splunk/ansible-role-for-splunk/issues/42

### DIFF
--- a/roles/splunk/handlers/main.yml
+++ b/roles/splunk/handlers/main.yml
@@ -77,6 +77,14 @@
     state: restarted
   become: true
 
-- name: restart auditd service
+- name: restart redhat auditd service
   command: service auditd condrestart
   become: true
+  when: ansible_os_family == 'RedHat'
+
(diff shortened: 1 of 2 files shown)


mason-splunk:
`roles/splunk/templates/genpstacks.sh.j2  100755 → 100644` I don't think changes to this file should be in this PR

arengifoc:
> 
> 
> `roles/splunk/templates/genpstacks.sh.j2 100755 → 100644` I don't think changes to this file should be in this PR

You're right. That was my mistake. I've fixed the commit and it doesn't include that genpstacks.sh.j2 file anymore. Take a look at it please.

mason-splunk (approved):

Issue #27, opened 2021-02-23
Bug: Idempotency for user creation .bashrc/.bash_profile


mason-splunk:
Two tasks report changes even when the files have not actually changed. Investigate cause of this and remediate to ensure role is idempotent.

1. splunk user creation
2. configuration of .bashrc and .bash_profile

mason-splunk:
Resolved in https://github.com/splunk/ansible-role-for-splunk/releases/tag/v1.2.9

PR #44, opened 2021-03-04
#27 - Correction of the Idempotency of the tasks

Jalkar:
Bugfix of issue #27 
Correction of the idempotency based on the following findings :
- https://symfonycasts.com/screencast/ansible/idempotency-changed-when (for .bashrc & .bashprofile)
- https://nickjanetakis.com/blog/debugging-an-idempotency-related-user-group-issue-with-ansible (for user creation)

### DIFF
--- a/roles/splunk/tasks/configure_bash.yml
+++ b/roles/splunk/tasks/configure_bash.yml
@@ -10,3 +10,4 @@
   loop:
     - { bashtemplate: 'bashrc.j2', bashfilepath: '.bashrc' }
     - { bashtemplate: 'bash_profile.j2', bashfilepath: '.bash_profile' }
+  changed_when: False
--- a/roles/splunk/tasks/configure_bash.yml
+++ b/roles/splunk/tasks/configure_bash.yml
@@ -10,4 +10,3 @@
   loop:
     - { bashtemplate: 'bashrc.j2', bashfilepath: '.bashrc' }
     - { bashtemplate: 'bash_profile.j2', bashfilepath: '.bash_profile' }
-  changed_when: False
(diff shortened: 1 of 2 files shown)


roles/splunk/tasks/install_splunk.yml
```diff
@@ -13,6 +13,7 @@
         name: "{{ splunk_nix_user }}"
         groups: "{{ splunk_nix_group }}"
         home: "{{ splunk_home }}"
+        append: true
```
mason-splunk, line 16: Good catch

roles/splunk/tasks/configure_bash.yml
```diff
@@ -10,3 +10,4 @@
   loop:
     - { bashtemplate: 'bashrc.j2', bashfilepath: '.bashrc' }
     - { bashtemplate: 'bash_profile.j2', bashfilepath: '.bash_profile' }
+  changed_when: False
```
mason-splunk, line 13: This doesn't really fix idempotency for this task. It just tells Ansible to report that nothing changed even if it did make changes to these files.

mason-splunk: I tested solely the `append: true` part of this PR without the change to `configure_bash.yml` and was able to confirm that it resolves the idempotency issue for both tasks. Please remove this change from `configure_bash.yml` and we'll get your PR merged. Thank you!

Jalkar: oups, it tested but obviously i'm unable to read "OK" vs "CHANGED" 👎 
I'll update the PR

Jalkar: And in fact without my changes, i don't reproduce the issue for the `configure_bash.yml`
First execution : 
image

Second execution : 
image

Issue #50, opened 2021-03-13
Enhancement: Add support for splunk.secret to be passed as a variable


chrisbalmer:
I'd like to be able to pass the splunk.secret as a variable instead of a file so it can be pulled from a secret storage (i.e. HashiCorp Vault, 1Password, etc).

shparker1977:
You can also do this with the builtin ansible copy module:

```
- name: Push the secrets file
  tags: install
  copy:
    content: '{{ splunk_secret' }}'
    dest: '{{ splunk_home }}/etc/auth/splunk.secret'
  notify:
    - restart splunk
  ```

peterchenadded:
Other code needs to be updated, otherwise all pass4symmkey and sslPasswords won't be able to decrypt and your server probably won't start.

PR #51, opened 2021-03-13
Added support for splunk.secret to be passed as a variable

chrisbalmer:
This is a PR to resolve issue #50 

Added a new task to setup the splunk.secret file as a template if the `splunk_secret_text` variable is set. I also included an example script to generate a new secret to store in a secret vault using the Splunk UF docker image.


roles/splunk/defaults/main.yml
```diff
@@ -20,7 +20,8 @@ clientName: undefined
 splunk_admin_username: admin
 splunk_admin_password: undefined # Use ansible-vault encrypt_string, e.g. ansible-vault encrypt_string --ask-vault-pass 'var_value_to_encrypt' --name 'var_name'
 splunk_configure_secret: false # If set to true, you need to update files/splunk.secret
-splunk_secret_file: splunk.secret # Used to specify your splunk.secret filename(s), files should be placed in the "files" folder of the role
+#splunk_secret_file: splunk.secret # Used to specify your splunk.secret filename(s), files should be placed in the "files" folder of the role
```
mason-splunk, line 23: Need to avoid having variables commented out in the default variables file.

scripts/generate-secrets.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
```
mason-splunk, line 1: This file doesn't belong in the role

roles/splunk/tasks/configure_splunk_secret.yml
```diff
@@ -7,4 +7,16 @@
     mode: 0644
   become: true
   notify: restart splunk
-  when: splunk_configure_secret
+  when: splunk_configure_secret and splunk_secret_file is defined
+
+- name: Install splunk.secret
```
mason-splunk, line 12: We should differentiate between the tasks in `name` if we're going to provide to multiple implementation methods. e.g. Line 1 "Install splunk.secret from file" and line 12 "Install splunk.secret from text"

roles/splunk/defaults/main.yml
mason-splunk: Please follow the existing pattern - i.e. default value of "undefined" that is checked in the task. This has the advantage of listing all valid variables in the default vars, but still not implementing them unless they've been configured via group_vars/host_vars/playbook vars.

mason-splunk (changes requested):

scripts/generate-secrets.sh
chrisbalmer: Script removed

roles/splunk/defaults/main.yml
chrisbalmer: Fixed to match the pattern. I set both to undefined. Let me know if you want the file one returned to `splunk.secret`.

roles/splunk/tasks/configure_splunk_secret.yml
chrisbalmer: Added names matching your request so the tasks are different.

roles/splunk/tasks/configure_splunk_secret.yml
```diff
@@ -7,4 +7,16 @@
     mode: 0644
   become: true
   notify: restart splunk
-  when: splunk_configure_secret
+  when: splunk_configure_secret and splunk_secret_file is defined
```
mason-splunk, line 10: This needs to be changed to:
```
  when: 
    - splunk_configure_secret
    - splunk_secret_file != 'undefined'
 ```

File: roles/splunk/tasks/configure_splunk_secret.yml [hunk 5]
```diff
@@ -7,4 +7,16 @@
...
+- name: Install splunk.secret from text
+  template:
+    src: splunk.secret.j2
+    dest: "{{ splunk_home }}/etc/auth/splunk.secret"
+    owner: "{{ splunk_nix_user }}"
+    group: "{{ splunk_nix_group }}"
+    mode: 0644
+  become: true
+  notify: restart splunk
+  when: splunk_configure_secret and splunk_secret_text is defined
```
mason-splunk, line 21: This needs to be changed to:
```
  when: 
    - splunk_configure_secret
    - splunk_secret_text != 'undefined'
 ```

mason-splunk (changes requested):
Almost there, but `is defined` in the `when` only checks if the variable is actually defined anywhere with a value. Since we are setting a default value with the string value "undefined" in defaults/main.yml, this will always evaluate to true with your current conditionals.

roles/splunk/tasks/configure_splunk_secret.yml
chrisbalmer: Good catch, updated.

roles/splunk/tasks/configure_splunk_secret.yml
chrisbalmer: Good catch, updated.

PR #52, opened 2021-03-17
Added web proxy support

ForsetiJan:
Added support for configuring the http(s) proxy.


roles/splunk/tasks/configure_web_proxy.yml
```diff
@@ -0,0 +1,43 @@
+- name: create the app path
+  file:
+    path: "{{ splunk_home }}/etc/system/local"
```
mason-splunk, line 3: I don't think this task is required.

mason-splunk:
@ForsetiJan  This could be accomplished by using `configure_apps.yml` to deploy an app/git repo containing server.conf with the proxy configurations. Can you explain why your use case needs to have this configuration implemented via Ansible vs. the existing supported method of deploying an app that contains the server.conf configuration?

ForsetiJan:
@mason-splunk  That actually makes total sense! Let's not merge

Issue #57, opened 2021-03-19
Issue : main.yml is reloading systemctl even if we use init.d


Jalkar:
May be i've done something wrong on my box, but for me if I use init.d i should not see a trigger on systemctl daemon reload

solution trigger the reload only if we are not using initd
```yaml
- name: reload systemctl daemon
  systemd:
    daemon_reload: true
  become: true
  when: not splunk_use_initd
```

mason-splunk:
We did some testing on this. What we found is that while the changes do take effect without reloading systemd, systemd will complain when the service is started if you do not reload it after making changes to init.d files:
```
# service splunk start
Starting splunk (via systemctl):  Warning: splunk.service changed on disk. Run 'systemctl daemon-reload' to reload units.
# echo $?
0
```
Although the return code is success (0) we will keep the reload to avoid this warning message.

Jalkar:
This implies that you have systemd installed on your machine :)
My tests was running on an Alpine linux which did not ship with systemd (hence the issue)

Possible workaround is to ignore error but i think it might miss some "real" error in the process

mason-splunk:
Alpine Linux is not supported by this role. Please see [main.yml in meta/](https://github.com/splunk/ansible-role-for-splunk/blob/master/roles/splunk/meta/main.yml) for a list of supported operating systems.

PR #58, opened 2021-03-19
Do not reload systemd when using init.d

Jalkar:
issue #57 - Do not reload systemd when using init.d


mason-splunk (changes requested):
This PR contains commits with additional changes beyond the scope of the PR title/description.

mason-splunk:
The current behavior is correct. Refer to https://github.com/splunk/ansible-role-for-splunk/issues/57 for details.

Issue #49, opened 2021-03-11
configure_apps.yml does not support a git repo where there is multiple apps


Jalkar:
On my gitlab, i've a repo name `splunk_apps` it contains all the apps for the same "project"!
```
- app_1
- app_2
- ...
```

May be i've missed something in my configuration but it does not seems possible to install only app_1 on a server and app_2 on a second

For me the issue lies in the task `Download defined Git repos to local Ansible host`

```
    - name: Download defined Git repos to local Ansible host
      git:
        accept_hostkey: true
        repo: "{{ item.git_server | default(git_server) }}/{{ item.git_project | default(git_project) }}/{{ item.name }}"
        version: "{{ item.git_version | default(git_version) }}"
        dest: "{{ git_local_clone_path }}{{ ansible_nodename }}/{{ item.name }}"
        key_file: "{{ git_key }}"
        force: true
      loop: "{{ git_apps }}"
      delegate_to: localhost
      changed_when: false
      check_mode: false
```

it's looping on the param `git_apps` and is expecting each of them to be a standalone git repo. For me each standalone repo should be configured by the param `git_project` only

Did I miss something ?  May be we can add the 2 possilities with a new parameter `git_multiple_app_per_repo`

mason-splunk:
Possible implementation: Within git_apps vars, have an optional variable where the user can list out the specific folders within the repository that they would like deployed (default to all). 

Notes: 
- The installation path would likely need to be configurable at the folder level as well
- Still need to accommodate different handlers as well
- Need to evaluate further if this change would break the rsync deployment method
- This feature may need to be added at the same time as https://github.com/splunk/ansible-role-for-splunk/issues/48

Jalkar:
We tend to use 1 git repo for all roles. This is to simplified the dev process & avoid duplicates or confusion (i can see you props.conf)

On example is : 
- an app with sourcetype info
- an app with inputs script
- an app with dashboards

Some of them will need to be deployed to a HF and the other to a single instance. 

we tend to have all the app on the same level (some times in a subfolder `splunk_apps` or directly on the root dir of the git repo

ForsetiJan:
I resolved this challenge by creating a new task called install_multi_app.yml.
It will still take the git_apps variable to run call configure_apps.yml

The determine_handler.yml task is the following:


```
- name: sync the apps repo
  include_tasks: configure_apps.yml

[...]

PR #62, opened 2021-03-31
my attempts to install multiple apps from git repo

Jalkar:
My proposition to solve #49

The idea is to use the existing parameters
```
git_server : the git_server "group" project (for example github.com/splunk)
git_project : the "real" project containing the app "ansible-role-for-splunk"
git_apps : the list of app names to be isntalled (the folder name in fact)
```

The catch to install "all" apps is to not set a git_apps.name in the inventory : 

```
git_apps:
    - name: ""
```
will install everything

### DIFF
--- a/roles/splunk/tasks/configure_apps.yml
+++ b/roles/splunk/tasks/configure_apps.yml
@@ -42,12 +42,11 @@
     - name: Download defined Git repos to local Ansible host
       git:
         accept_hostkey: true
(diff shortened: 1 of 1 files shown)


Jalkar:
this is crap -_-'

mason-splunk:
We'll work on adding this for you @Jalkar  just give us some time. :)

Jalkar:
Thanks ^^

In fact the change I was attempting to push only worked on one of my specific use cases this is why I called it crap a delete it :D

PR #64, opened 2021-04-26
Suggest extended handling of licenses

schneewe:
I may suggest a little change in the license handling.
The licensemaster should not set the license master uri, it should install the license instead.

So we can define a list of license files in the role/file directory stored in the splunk_licenses variable and set them for the licensemaster (in which way someone prefer)

Error handling of adding licenses via CLI can be ignored, however we need to define when it is changed, so splunk can restart.

### DIFF
--- a/roles/splunk/defaults/main.yml
+++ b/roles/splunk/defaults/main.yml
@@ -13,7 +13,7 @@ splunk_install_type: undefined # There are two ways to configure this. The easie
 splunk_install_path: /opt # Base directory on the operating system to which splunk should be installed
 splunk_nix_user: splunk
 splunk_nix_group: splunk
-splunk_licenses: undefined
+splunk_licenses: []
 splunk_uri_lm: undefined
 splunk_uri_cm: undefined
 splunk_uri_ds: undefined # e.g. mydeploymentserver.mydomain.com:8089 ; Note that you must also configure the clientName var under either group_vars or host_vars for deploymentclient.conf to be configured
--- a/roles/splunk/defaults/main.yml
+++ b/roles/splunk/defaults/main.yml
@@ -13,6 +13,7 @@ splunk_install_type: undefined # There are two ways to configure this. The easie
 splunk_install_path: /opt # Base directory on the operating system to which splunk should be installed
 splunk_nix_user: splunk
 splunk_nix_group: splunk
+splunk_licenses: undefined
 splunk_uri_lm: undefined
 splunk_uri_cm: undefined
 splunk_uri_ds: undefined # e.g. mydeploymentserver.mydomain.com:8089 ; Note that you must also configure the clientName var under either group_vars or host_vars for deploymentclient.conf to be configured
--- a/roles/splunk/tasks/configure_license.yml
+++ b/roles/splunk/tasks/configure_license.yml
@@ -1,31 +1,34 @@
 ---
 - block:
-  - name: Copy Licenses to Master
+  - name: Copy licenses to the licensemaster
     copy:
-      src: "{{ item }}"
-      dest: "/tmp/{{ item }}"
+      src: "{{ item.src }}"
+      dest: "{{ item.dest }}"
+      owner: "{{ splunk_nix_user }}"
+      group: "{{ splunk_nix_group }}"
+      mode: 0644
     loop: "{{ splunk_licenses }}"
     
-  - name: Install license on licensemaster
-    command: "{{ splunk_home }}/bin/splunk add licenses /tmp/{{ item }} -auth {{ splunk_auth }}"
+  - name: Install licenses on the licensemaster
+    command: "{{ splunk_home }}/bin/splunk add licenses {{ item.dest }} -auth {{ splunk_auth }}"
     become: true
     become_user: "{{ splunk_nix_user }}"
     loop: "{{ splunk_licenses }}"
     ignore_errors: yes
+    no_log: true
     register: license_result
-    changed_when: '"The licenses object has been added" in license_result.stdout'
+    changed_when: license_result.rc == 0
     notify: restart splunk
 
-  - name: Clean Up licenses
+  - name: Cleanup licenses on the licensemaster that were copied
     file:
-      path: "/tmp/{{ item }}"
+      path: "{{ item.dest }}"
       state: absent
     become: true
     loop: "{{ splunk_licenses }}"
 
   when:
     - "'licensemaster' in group_names"
-    - splunk_licenses is defined
     - splunk_licenses != 'undefined'
 
 - name: Set license master_uri
--- a/roles/splunk/tasks/configure_license.yml
+++ b/roles/splunk/tasks/configure_license.yml
@@ -2,31 +2,13 @@
 - block:
   - name: Copy licenses to the licensemaster
     copy:
-      src: "{{ item.src }}"
-      dest: "{{ item.dest }}"
+      src: "{{ item }}"
+      dest: "{{ splunk_home}}/etc/license/enterprise/{{ item }}"
       owner: "{{ splunk_nix_user }}"
       group: "{{ splunk_nix_group }}"
       mode: 0644
     loop: "{{ splunk_licenses }}"
-    
-  - name: Install licenses on the licensemaster
-    command: "{{ splunk_home }}/bin/splunk add licenses {{ item.dest }} -auth {{ splunk_auth }}"
-    become: true
-    become_user: "{{ splunk_nix_user }}"
-    loop: "{{ splunk_licenses }}"
-    ignore_errors: yes
-    no_log: true
-    register: license_result
-    changed_when: license_result.rc == 0
     notify: restart splunk
-
-  - name: Cleanup licenses on the licensemaster that were copied
-    file:
-      path: "{{ item.dest }}"
-      state: absent
-    become: true
-    loop: "{{ splunk_licenses }}"
-
   when:
     - "'licensemaster' in group_names"
     - splunk_licenses != 'undefined'
--- a/roles/splunk/tasks/configure_license.yml
+++ b/roles/splunk/tasks/configure_license.yml
@@ -1,4 +1,33 @@
 ---
+- block:
+  - name: Copy Licenses to Master
+    copy:
+      src: "{{ item }}"
+      dest: "/tmp/{{ item }}"
+    loop: "{{ splunk_licenses }}"
+    
+  - name: Install license on licensemaster
+    command: "{{ splunk_home }}/bin/splunk add licenses /tmp/{{ item }} -auth {{ splunk_auth }}"
+    become: true
+    become_user: "{{ splunk_nix_user }}"
+    loop: "{{ splunk_licenses }}"
+    ignore_errors: yes
+    register: license_result
+    changed_when: '"The licenses object has been added" in license_result.stdout'
+    notify: restart splunk
+
+  - name: Clean Up licenses
+    file:
+      path: "/tmp/{{ item }}"
+      state: absent
+    become: true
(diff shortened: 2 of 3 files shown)


roles/splunk/tasks/configure_license.yml
```diff
@@ -1,4 +1,33 @@
...
+  - name: Clean Up licenses
+    file:
+      path: "/tmp/{{ item }}"
+      state: absent
+    become: true
+    loop: "{{ splunk_licenses }}"
+
+  when:
+    - "'licensemaster' in group_names"
+    - splunk_licenses is defined
```
mason-splunk, line 28: This line isn't needed because you are already defining `splunk_licenses` in defaults/main.yml

roles/splunk/tasks/configure_license.yml
```diff
@@ -1,4 +1,33 @@
...
+    command: "{{ splunk_home }}/bin/splunk add licenses /tmp/{{ item }} -auth {{ splunk_auth }}"
+    become: true
+    become_user: "{{ splunk_nix_user }}"
+    loop: "{{ splunk_licenses }}"
+    ignore_errors: yes
+    register: license_result
+    changed_when: '"The licenses object has been added" in license_result.stdout'
+    notify: restart splunk
+
+  - name: Clean Up licenses
```
mason-splunk, line 19: Please change to `  - name: Clean up licenses` to be consistent with the rest of the role

roles/splunk/tasks/configure_license.yml
```diff
@@ -1,4 +1,33 @@
 ---
+- block:
+  - name: Copy Licenses to Master
```
mason-splunk, line 3: Please change to `  - name: Copy licenses to master` to be consistent with the rest of the role

roles/splunk/tasks/configure_license.yml
```diff
@@ -1,4 +1,33 @@
...
+    copy:
+      src: "{{ item }}"
+      dest: "/tmp/{{ item }}"
```
mason-splunk, line 6: Remove `/tmp/` as (1) it may not exist and (2) users can already enter a path when defining `splunk_licenses` so hardcoding it here isn't required

roles/splunk/tasks/configure_license.yml
```diff
@@ -1,4 +1,33 @@
...
+    loop: "{{ splunk_licenses }}"
+    
+  - name: Install license on licensemaster
+    command: "{{ splunk_home }}/bin/splunk add licenses /tmp/{{ item }} -auth {{ splunk_auth }}"
```
mason-splunk, line 10: Same thing here, please remove the hardcoded `/tmp/`

roles/splunk/tasks/configure_license.yml
```diff
@@ -1,4 +1,33 @@
...
+    become: true
+    become_user: "{{ splunk_nix_user }}"
+    loop: "{{ splunk_licenses }}"
+    ignore_errors: yes
+    register: license_result
+    changed_when: '"The licenses object has been added" in license_result.stdout'
```
mason-splunk, line 16: Please use `license_result.rc == 0` instead. The text could change in a future version, so return code is more future-proof.

roles/splunk/tasks/configure_license.yml
```diff
@@ -1,4 +1,33 @@
...
+    notify: restart splunk
```
mason-splunk, line 17: Please add a new line with `no_log: true` to avoid logging plaintext credentials from `splunk_auth`

roles/splunk/tasks/configure_license.yml
```diff
@@ -1,4 +1,33 @@
...
+      src: "{{ item }}"
+      dest: "/tmp/{{ item }}"
+    loop: "{{ splunk_licenses }}"
+    
+  - name: Install license on licensemaster
+    command: "{{ splunk_home }}/bin/splunk add licenses /tmp/{{ item }} -auth {{ splunk_auth }}"
+    become: true
+    become_user: "{{ splunk_nix_user }}"
+    loop: "{{ splunk_licenses }}"
+    ignore_errors: yes
```
mason-splunk, line 14: The play should fail if there is an error. Please remove `ignore_errors: yes`

roles/splunk/tasks/configure_license.yml
```diff
@@ -1,4 +1,33 @@
 ---
+- block:
+  - name: Copy Licenses to Master
+    copy:
```
mason-splunk, line 4: Ansible-lint will complain if you don't define ownership and mode. Please add the following 3 lines to your copy:
```
    owner: "{{ splunk_nix_user }}"
    group: "{{ splunk_nix_group }}"
    mode: 0644
```

roles/splunk/tasks/configure_license.yml
```diff
@@ -1,4 +1,33 @@
...
+    become_user: "{{ splunk_nix_user }}"
+    loop: "{{ splunk_licenses }}"
+    ignore_errors: yes
+    register: license_result
+    changed_when: '"The licenses object has been added" in license_result.stdout'
+    notify: restart splunk
+
+  - name: Clean Up licenses
+    file:
+      path: "/tmp/{{ item }}"
```
mason-splunk, line 21: Remove hardcoded `/tmp/`

mason-splunk (changes requested):

roles/splunk/tasks/configure_license.yml
schneewe: Yeah you totally right. Will change that

roles/splunk/tasks/configure_license.yml
schneewe: Yeah that would be a little bit tricky. 
If the license already exists we receive one, if the license expired we receive an error, if the license is invalid we receive one, but one of them is necessary to stop the play. As splunk can run without the license.
So the simplest way was to ignore errors, only changed is necessary, because we need to restart splunk.

roles/splunk/tasks/configure_license.yml
schneewe: In the current state, user is defining a source, so it has to exists on the ansible host (source) and on the licensemaster (target). Maybe that's not guaranteed at all.
So, what about to define splunk_licenses as hash table, so user can define where to store them temporary?

mason-splunk: You could just use two variables, one for `splunk_license_ansible_path` and one for `splunk_license_target_path`

roles/splunk/tasks/configure_license.yml
mason-splunk: Okay, this is good then.

roles/splunk/tasks/configure_license.yml
micahkemp: I think I'd agree with having the license be defined as a text blob in a variable, and installed via [the REST API](https://docs.splunk.com/Documentation/SplunkCloud/8.1.2103/RESTREF/RESTlicense#licenser.2Flicenses).  The `add licenses` command used later in this PR already requires auth, so there's nothing holding you back from making the change via authenticated REST API.

micahkemp: The difficulty may be in doing that idempotently, which is a requirement when working with Ansible.  The REST API returns the hash of the license, which is then used to query the REST API for the license details later.  What's not clear is if you could precalculate that, or if it's in the `.lic` file already (I didn't see it, but did see a `GUID`).

roles/splunk/tasks/configure_license.yml
micahkemp: I'm not a fan of ignoring errors as a means to accomplish idempotency.  Additional work should be done to make this process idempotent, whether by changing the method to use the API, or by parsing the output of the command for indication that it failed only because of a duplicate license.

roles/splunk/tasks/configure_license.yml
micahkemp: Another benefit of using the REST API, credentials aren't used somewhere they'd be logged.

roles/splunk/tasks/configure_license.yml
micahkemp: Regarding `/tmp`, you can create a tempfile/directory in ansible using builtin methods: https://docs.ansible.com/ansible/latest/collections/ansible/builtin/tempfile_module.html

roles/splunk/defaults/main.yml
```diff
@@ -13,6 +13,7 @@ splunk_install_type: undefined # There are two ways to configure this. The easie
 splunk_install_path: /opt # Base directory on the operating system to which splunk should be installed
 splunk_nix_user: splunk
 splunk_nix_group: splunk
+splunk_licenses: undefined
```
micahkemp, line 16: Wouldn't this make more sense to have a default value of `[]`?

roles/splunk/tasks/configure_license.yml
mason-splunk: @schneewe We think the best way to implement this functionality would actually be to use the `uri` module to send an HTTP POST to the API with the license in the payload. See: https://docs.splunk.com/Documentation/Splunk/8.1.3/RESTREF/RESTlicense#licenser.2Flicenses

You can reuse some of the parameters from https://github.com/splunk/ansible-role-for-splunk/blob/5163490b765a85c334fce1f4d6e316dbdb0f87f2/roles/splunk/tasks/install_apps.yml#L45  to handle authentication.

mason-splunk: To handle multiple licenses, the `splunk_licenses` var would contain a list of file paths leveraging the file_contents lookup. Then, use a loop with `splunk_licenses` in the `uri` module to set the payload.

schneewe: I get your point and completely agree, but I think about the very first run, this would not be possible because splunk isn't running therefore no API available. If splunk is already installed, this is the best solution.
@micahkemp the way with the temp folder is very nice, didn't know that.

micahkemp-splunk: @schneewe is there a reason the licenses need to be configured before first start of splunk?

mason-splunk: @micahkemp-splunk  I think with the current proposed implementation, it would be an order of operations thing, since configure_license.yml is called during the install process (since it previously only supported LM clients). That said, the easy way to fix that would be to just move these new tasks to their own file, like `configure_lm.yml` which could then be called as the next play after installation.

roles/splunk/defaults/main.yml
schneewe: Yeah thanks, you're right. I added this to the newest pull request.

schneewe:
I added a new suggest. We can copy the files to the licenses directory of Splunk. Restart of splunk can be handled by ansible.

roles/splunk/tasks/configure_license.yml
```diff
@@ -1,4 +1,18 @@
...
+      src: "{{ item }}"
+      dest: "{{ splunk_home}}/etc/license/enterprise/{{ item }}"
+      owner: "{{ splunk_nix_user }}"
+      group: "{{ splunk_nix_group }}"
+      mode: 0644
+    loop: "{{ splunk_licenses }}"
+    notify: restart splunk
+  when:
+    - "'licensemaster' in group_names"
+    - splunk_licenses != 'undefined'
```
mason-splunk, line 14: @schneewe Change this to `splunk_licenses|length > 0` or `splunk_licenses.0 is defined`

dtwersky:
@schneewe I created a separate PR that includes your code, and adds handling of more licensing configs

Issue #65, opened 2021-04-26
Feature: Renaming the Splunk groups with a prefix


schneewe:
I have a feature request (can make a PR too) to rename the groups in the inventory file (and tasks respectively) to a more splunk name, to prefix it with splunk_ 

I had a discussion with a customer and he had a good point. You can have multiple host in an inventory which are not related to splunk. As this is a role and you can have multiple of them, group names should not be that generic like "full" or "search", due to inventories are not role dependent.

mason-splunk:
This is a good suggestion. We can implement or if you want it sooner, feel free to PR.

schneewe:
I will do it, thank you.

schneewe:
I'm not sure what you mean. So, just having the current groups nested under a splunk group and limit a playbook run with the splunk group?

mason-splunk:
Yeah, like this:

```
all:
  children:

    splunk:
      children:
      
        full:
          children:

            licensemaster:
              hosts:
                my-lm-host:

        uf:
          children:

            east_coast:
              hosts:
                someforwarder

    some_other_application:
      children:
      ...
```

PR #69, opened 2021-04-29
Renamed Group names with prefix, Issue #65

schneewe:
As discussed in [#65 ](https://github.com/splunk/ansible-role-for-splunk/issues/65) I renamed the groups.
Addittionally I respect the new name of the manager node "clustermanager" and the old one, for migrating 
[Splunk Indexer Cluster](https://docs.splunk.com/Documentation/Splunk/8.1.3/InheritedDeployment/Deploymenttopology#Indexer_cluster)


schneewe:
Thanks @mason-splunk for the clarification, but I guess it's not worth to change the inventory file on this repo.
If you see it differently as it is a good point, I'll create a new PR.

Issue #18, opened 2021-02-12
Enhancement: Deploy an indexer cluster


mason-splunk:
https://docs.splunk.com/Documentation/Splunk/latest/Indexer/Clusterdeploymentoverview

PR #72, opened 2021-05-01
Add indexer clustering configurations

midnight5ky:
This pull request is to close out issue [#18](https://github.com/splunk/ansible-role-for-splunk/issues/18) - configure indexer cluster.

### DIFF
--- a/roles/splunk/tasks/configure_idxc_manager.yml
+++ b/roles/splunk/tasks/configure_idxc_manager.yml
@@ -28,7 +28,7 @@
     - { option: "replication_factor", value: "{{ splunk_idxc_rf }}" }
     - { option: "search_factor", value: "{{ splunk_idxc_sf }}" }
     - { option: "pass4SymmKey", value: "{{ splunk_idxc_key }}" }
-    - { option: "shcluster_label", value: "{{ splunk_idxc_label }}" }
+    - { option: "cluster_label", value: "{{ splunk_idxc_label }}" }
   when: splunk_v_num_only_url is version('8.1.0','<') or splunk_v_num_only is version('8.1.0','<')
 
 # Splunk version >= 8.1.X supports mode=manager
--- a/roles/splunk/tasks/configure_idxc_manager.yml
+++ b/roles/splunk/tasks/configure_idxc_manager.yml
@@ -8,7 +8,7 @@
 # https://docs.splunk.com/Documentation/Splunk/latest/Indexer/Configuremanagerwithserverconf
 - name: Setting clustering mode based on Splunk version number
   set_fact:
-    mode_value: "{% if splunk_v_num_only | int < 8.1 %}master{% else %}manager{% endif %}"
+    mode_value: "{% if splunk_version_release | int < 8.1 %}master{% else %}manager{% endif %}"
 
 - name: Configure clustering stanza for cluster manager node
   ini_file:
--- a/roles/splunk/tasks/configure_idxc_manager.yml
+++ b/roles/splunk/tasks/configure_idxc_manager.yml
@@ -1,4 +1,5 @@
 ---
+# https://docs.splunk.com/Documentation/Splunk/latest/Indexer/Configuremanagerwithserverconf
 - name: Configure clustering stanza for cluster manager node
   ini_file:
     path: "{{ splunk_home }}/etc/system/local/server.conf"
--- a/roles/splunk/tasks/configure_idxc_manager.yml
+++ b/roles/splunk/tasks/configure_idxc_manager.yml
@@ -1,6 +1,39 @@
 ---
+- name: Run splunk version command to check currently installed version
+  command: "{{ splunk_home }}/bin/splunk version --answer-yes --auto-ports --no-prompt --accept-license"
+  register: current_version
+  become: true
+  become_user: "{{ splunk_nix_user }}"
+  changed_when: false
+
+- name: Set splunk version number
+  set_fact:
+    splunk_v_num_only: "{{ current_version.stdout | regex_search ('\\d+\\.\\d+\\.\\d+') }}"
+
+# Splunk version < 8.1.X supports mode=master
+# https://docs.splunk.com/Documentation/Splunk/8.0.9/Indexer/Configuremasterwithserverconf
+- name: Configure clustering stanza for cluster manager node on Splunk version < 8.1.X
+  ini_file:
+    path: "{{ splunk_home }}/etc/system/local/server.conf"
+    section: clustering
+    option: "{{ item.option }}"
+    value: "{{ item.value }}"
+    mode: 0644
+    owner: "{{ splunk_nix_user }}"
+    group: "{{ splunk_nix_group }}"
+  become: true
+  notify: restart splunk
+  loop:
+    - { option: "mode", value: "master" }
+    - { option: "replication_factor", value: "{{ splunk_idxc_rf }}" }
+    - { option: "search_factor", value: "{{ splunk_idxc_sf }}" }
+    - { option: "pass4SymmKey", value: "{{ splunk_idxc_key }}" }
+    - { option: "shcluster_label", value: "{{ splunk_idxc_label }}" }
+  when: splunk_v_num_only_url is version('8.1.0','<') or splunk_v_num_only is version('8.1.0','<')
+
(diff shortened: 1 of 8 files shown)


schneewe:
The new values of manager and peer for mode option are only valid for Splunk 8.1.x and higher. An additional check would be great which version will be installed or upgraded to.

midnight5ky:
@schneewe - Thanks for pointing that out - I'll work on something to dynamically set the mode based on the Splunk version the host is running.

midnight5ky:
Just a quick note - even setting the "mode=peer" in the CLI for the IDXC member, the server.conf still shows as "mode=slave" under the `[cluster]` stanza.

midnight5ky:
I noticed another issue where the `notify: restart splunk` won't work if you run check_splunk.yml and then configure_idxc_*.yml tasks - due to how "restart splunk" handler is configured:

```
- name: restart splunk
  service:
    name: "{{ splunk_service }}"
    state: restarted
  become: true
  when: not start_splunk_handler_fired
```

I had to specifically put in restart splunk task in the task playbook so that the CM/IDX can restart.

```
- name: restart splunk
  service:
    name: "{{ splunk_service }}"
    state: restarted
  become: true
```

schneewe:
if your configure_idxc_*.yml tasks are the last tasks running, you can call the meta module "flush_handlers" before, then the notify calls get executed and then your normal task should work.

roles/splunk/tasks/configure_idxc_manager.yml
```diff
@@ -0,0 +1,53 @@
+---
+- name: Run splunk version command to check currently installed version
```
mason-splunk, line 2: Don't duplicate existing tasks. If you want to re-use a task, split it out and use `include_tasks` instead.

mason-splunk (changes requested):
Please remove duplicated code.

roles/splunk/tasks/configure_idxc_manager.yml
midnight5ky: Gotcha, thanks Mason - I'll work on splitting that task out.

midnight5ky:
@schneewe 

> if your configure_idxc_*.yml tasks are the last tasks running, you can call the meta module "flush_handlers" before, then the notify calls get executed and then your normal task should work.

Thanks for the suggestion - I was brainstorming of a way to reset the handler yesterday but couldn't get to it. I'll definitely use your suggestion!

roles/splunk/vars/main.yml
```diff
@@ -2,6 +2,7 @@
 # Automatically determine desired splunk version string
 splunk_file: "{{ splunk_package_url | regex_search('(splunk(?:forwarder)?\\-\\d+[^/]*\\.tgz)') }}"
 splunk_v: "{{ splunk_package_url | regex_search('(\\d+\\.\\d+\\.\\d+(?:\\.\\d+)?)') }}"
+splunk_v_num_only_url: "{{ splunk_package_url | regex_search('(\\d+\\.\\d+\\.\\d+)') }}"
```
mason-splunk, line 5: You're running the splunk version command to get the current version, so there's no need to also get the version from the url.

midnight5ky: Oh yeah - you're right. I just included that there in case people were running check_splunk.yml at first. Don't know what I was thinking - I'll remove that.

roles/splunk/tasks/configure_idxc_manager.yml
```diff
@@ -0,0 +1,53 @@
...
+  command: "{{ splunk_home }}/bin/splunk version --answer-yes --auto-ports --no-prompt --accept-license"
+  register: current_version
+  become: true
+  become_user: "{{ splunk_nix_user }}"
+  changed_when: false
+
+- name: Set splunk version number
```
mason-splunk, line 9: Move this to the version check task as well. Also, I think you should only be checking the major and minor version (X.Y).

roles/splunk/tasks/configure_idxc_manager.yml
```diff
@@ -0,0 +1,53 @@
...
+  set_fact:
+    splunk_v_num_only: "{{ current_version.stdout | regex_search ('\\d+\\.\\d+\\.\\d+') }}"
+
+# Splunk version < 8.1.X supports mode=master
+# https://docs.splunk.com/Documentation/Splunk/8.0.9/Indexer/Configuremasterwithserverconf
+- name: Configure clustering stanza for cluster manager node on Splunk version < 8.1.X
```
mason-splunk, line 15: Instead of duplicating this task for different versions, use `set_fact` to set appropriate values for a new var `mode_value` based on the version. Then, you can do all the server.conf configuration in one task.

roles/splunk/tasks/configure_idxc_manager.yml
midnight5ky: Good point - Makes much better sense since only version >= 8.1 has the updated clustering mode values.

roles/splunk/tasks/configure_idxc_member.yml
```diff
@@ -0,0 +1,37 @@
...
+  changed_when: false
+
+- name: Set splunk version number
+  set_fact:
+    splunk_v_num_only: "{{ current_version.stdout | regex_search ('\\d+\\.\\d+\\.\\d+') }}"
+
+# Splunk version < 8.1.X supports mode=slave
+# https://docs.splunk.com/Documentation/Splunk/8.0.9/Indexer/ConfigurepeerswithCLI
+- name: Configure idxc member
+  command: "{{ splunk_home }}/bin/splunk edit cluster-config -mode slave -auth {{ splunk_auth }} -master_uri {{ splunk_uri_cm }} -replication_port {{ splunk_idxc_rep_port }} -secret {{ splunk_idxc_key }}"
```
mason-splunk, line 16: Do the same thing as server.conf on this one. Don't duplicate the task.

roles/splunk/tasks/configure_idxc_manager.yml
mason-splunk: I'd recommend moving https://github.com/splunk/ansible-role-for-splunk/blob/master/roles/splunk/tasks/check_splunk.yml#L21 to a new task called `check_splunk_version.yml` then include that where needed.

roles/splunk/tasks/configure_idxc_manager.yml
```diff
@@ -0,0 +1,36 @@
...
+# https://docs.splunk.com/Documentation/Splunk/8.0.9/Indexer/Configuremasterwithserverconf
+# Splunk version >= 8.1 supports mode=manager
+# https://docs.splunk.com/Documentation/Splunk/latest/Indexer/Configuremanagerwithserverconf
+- name: Setting clustering mode based on Splunk version number
+  set_fact:
+    mode_value: {% if splunk_v_num_only < 8.1 %}'master'{% else %}'manager'{% endif %}
+
+- name: view mode_value
+  debug:
+    msg: "mode value is {{ mode_value }}"
```
mason-splunk, line 15: Remove this

midnight5ky: Oh shoot, thanks - forgot to remove after testing.

mason-splunk (approved):

mason-splunk (approved):

roles/splunk/tasks/configure_idxc_manager.yml
```diff
@@ -0,0 +1,32 @@
...
+  become: true
+  notify: restart splunk
+  loop:
+    - { option: "mode", value: "{{ mode_value}}" }
+    - { option: "replication_factor", value: "{{ splunk_idxc_rf }}" }
+    - { option: "search_factor", value: "{{ splunk_idxc_sf }}" }
+    - { option: "pass4SymmKey", value: "{{ splunk_idxc_key }}" }
+    - { option: "cluster_label", value: "{{ splunk_idxc_label }}" }
+
+- name: Force handlers to run
```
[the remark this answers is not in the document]
midnight5ky: Gotcha - I removed it here: https://github.com/splunk/ansible-role-for-splunk/pull/72/commits/05a95cb5408a62c1898c5ab37a9c249ba4002262

I think you're looking at an older version of that.

Issue #68, opened 2021-04-29
Swap 'undefined' for '~' to denote initial null values


mason-splunk:
Use `<varname> is defined` instead of checking for a literal string value of 'undefined'. Functionally it works the same, but it's a little cleaner.

mason-splunk:
@flowerysong pointed out in https://github.com/splunk/ansible-role-for-splunk/pull/74 that this suggested implementation will not work. We could set default values to `""` and then check if the values are truthy, but that is only supported in Ansible 2.10+ (see https://docs.ansible.com/ansible/latest/user_guide/playbooks_tests.html) so the role would break for users that are on an older version. We will leave the current implementation as-is.

PR #74, opened 2021-05-05
Swap undefined string to tilde character

midnight5ky:
This pull request is to swap 'undefined' string to tilde character to define null value; relates to issue https://github.com/splunk/ansible-role-for-splunk/issues/68.


flowerysong:
This will not work as you intend it to; defining a variable as null is still a definition, so `is defined` will always be true. You need to either leave the variable entirely undefined or check its value (like the current code does.)

```yaml
- hosts: localhost
  gather_facts: false
  vars:
    foo: ~
    bar: false
  tasks:
    - debug:
        msg: |
          foo: {{ foo is defined }} / {{ foo != None }}
          bar: {{ bar is defined }} / {{ bar != None }}
          baz: {{ baz is defined }}
```

Output:
```
TASK [debug] *******************************************************************
ok: [localhost] => 
  msg: |-
    foo: True / False
    bar: True / True
    baz: False
```

mason-splunk:
Closing this out as we've decided not to move forward with this change. Refer to https://github.com/splunk/ansible-role-for-splunk/issues/68 for details.

Issue #108, opened 2021-10-22
splunk_version is not set up correctly when splunk_package_url_full contains IP address


karbovski:
```
TASK [../roles/splunk : Checkpoint: Version] **********************************************************************************************************************************************************************
ok: [cm01] => {
    "msg": "The value of splunk_version is: Splunk 10.10.10.10 (build ae6821b7c64b) and the current_version is: Splunk 8.2.2.1 (build ae6821b7c64b)"
```

when variable **splunk_package_url_full** is URL with IP address.

`splunk_package_url_full: http://10.10.10.10/files/splunk-8.2.2.1-ae6821b7c64b-Linux-x86_64.tgz`

The consequence of that is that Splunk will be reinstalled every time check_splunk.yml. The role is not idempotent anymore.

PR #112, opened 2021-12-23
#108 - Fix Regex

Jalkar:
Fix Regex for the version extraction when IP are in the provided url


dtwersky:
This is no longer valid since the version is determined with the `splunk_package_version` variable [here](https://github.com/splunk/ansible-role-for-splunk/blob/master/roles/splunk/defaults/main.yml#L9)

PR #119, opened 2022-06-01
Added sysctl to allow read of dmesg

jewnix:
Some distros have `kernel.dmesg_restrict` set to `1` by default. This prevents the Splunk_TA_nix app from running the `dmesg` command. This add the ability change that setting.

The `procps` needs to be restarted for the change to take effect.


roles/splunk/tasks/configure_os.yml
```diff
@@ -16,3 +16,7 @@
 - name: Disable THP
   include_tasks: configure_thp.yml
   when: "'full' in group_names"
+
+- name: Enable read for dmesg
+  include_tasks: configure_dmesg.yml
+  when: configure_dmesg == true
```
anelson-splunk, line 22: Is a comparison really necessary here? Or can you just set `when: config_dmesg` ?

anelson-splunk (approved):

micahkemp-splunk (approved):

PR #120, opened 2022-07-21
Add more licensing options

jewnix:
Redesign of the logic of how licenses are configured.
This allows adding license files, change an instance from being a peer to a localslave or set an instance to use the built-in licenses that ship with the application.
Added the option to used `manager_uri` instead of `master_uri` for versions running version 9.0+
The splunk version check now runs in `--check` mode. This prevents checks from failing.


jewnix:
There are multiple features being added, so I will combine multiple PR in to one, in separate commits.

PR #121, opened 2022-07-21
Added config to disable THP with tuned

jewnix:
The current method of disabling THP is a bit of a hack. Large Database vendors (e.g. Oracle) use `tuned` to apply performance tuning to their applications.


jewnix:
There are multiple features being added, so I will combine multiple PR in to one, in separate commits.

PR #122, opened 2022-07-26
Multiple updates

jewnix:
There are a few features/fixes/enactments here
* Complete rewrite of the `configure_license.yml`. You can now specify built-in licenses (e.g Forwarder), license files or license peer.
* Allow installation of one or more license files.
* Allow use of `tuned` to disable THP.
* Package will no longer downgrade if a newer version is installed, but the available version is older.
* `splunk_version_release` now has the full release version instead of only the major and minor. This allows us to reliably check if the available package is newer than the one that is currently installed.
* Updated to version 9.0.0.1
* You can now specify a specific version and build number in `host_vars` or `group_vars` with the `splunk_package_version` and `build_id` variables, instead of the version being static across all hosts.
* Fixed conditionals in `configure_splunk_boot.yml` for systemd, which caused failure in some cases.
* `.gitignore` for license files and test environment.


anelson-splunk (approved):

midnight5ky (approved):

roles/splunk/tasks/configure_facl.yml
```diff
@@ -41,4 +46,6 @@
         - restart non-redhat auditd service
       when: result_auditd_conf.stat.exists
 
-  when: splunk_nix_user != 'root'
+  when:
+    - least_privileged is not defined or "'full' in group_names"
```
dtwersky, line 50: As of version 9.0.0.x, least_privileged mode is not yet available on full installation. This may change in the future

dtwersky (approved):

PR #126, opened 2022-08-17
fix systemd

red667:
Since the splunk binary creates a systemd file with the wrong CGROUPS I added this small fix to the configure_systemd.yml. It will edit the correct value to the ExecStartPost option of the splunk systemd file.


roles/splunk/tasks/configure_systemd.yml
```diff
@@ -14,4 +14,5 @@
     - { option: "LimitDATA", value: "infinity" }
     - { option: "LimitCORE", value: "infinity" }
     - { option: "TasksMax", value: "infinity" }
+    - { option: "ExecStartPost", value: "/bin/bash -c \"chown -R splunk:splunk /sys/fs/cgroup/system.slice/%n\"" }  # Systemd CGROUP Fix
```
dtwersky, line 17: What OS and version did you see this? I looked on a Ubuntu 20.04 and CentOS, and did not see that cgroup

red667: I verified the CGROUPS on CentOS Stream 9, Fedora 36 and Ubuntu 22.04.

Command to find the CGROUP:
```
$ sudo find /sys -name "system.slice*"
/sys/fs/cgroup/system.slice
```
I'm running som Splunk POCs on CentOS Stream 9 and the systemd service never worked. Some researching led me to these post(s):
https://community.splunk.com/t5/Installation/Why-is-Systemd-broken-on-new-install/m-p/482885#M10392

red667: Redhat 9 should also have it like this:
https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/9/html/managing_monitoring_and_updating_the_kernel/assembly_using-systemd-to-manage-resources-used-by-applications_managing-monitoring-and-updating-the-kernel#proc_viewing-cgroups-of-processes_assembly_using-systemd-to-manage-resources-used-by-applications

dtwersky: This has not been tested on RHEL 9/CentOS Stream 9 or Ubuntu 22.04.
If we want this to work, that option should only be added if matches RHEL 9 etc. Can you add a conditional to add that line only if the version matches?

dtwersky: Also, what version of Splunk are you installing? I just tried the latest version of Splunk on 22.04, and it worked just fine.
```
# /opt/splunkforwarder/bin/splunk version
Splunk Universal Forwarder 9.0.0.1 (build 9e907cedecb1)

# cat /etc/lsb-release  | grep DISTRIB_RELEASE
DISTRIB_RELEASE=22.04

# uname -r
5.15.0-1017-aws

# find /sys -name "system.slice*"
/sys/fs/cgroup/memory/system.slice
/sys/fs/cgroup/blkio/system.slice
/sys/fs/cgroup/pids/system.slice
/sys/fs/cgroup/devices/system.slice
/sys/fs/cgroup/cpu,cpuacct/system.slice
/sys/fs/cgroup/systemd/system.slice
/sys/fs/cgroup/unified/system.slice

# systemctl cat SplunkForwarder.service 
# /etc/systemd/system/SplunkForwarder.service
#This unit file replaces the traditional start-up script for systemd
#configurations, and is used when enabling boot-start for Splunk on
#systemd-based Linux distributions.

[Unit]
Description=Systemd service file for Splunk, generated by 'splunk enable boot-start'
After=network-online.target
Wants=network-online.target

[Service]
Type=simple
Restart=always
ExecStart=/opt/splunkforwarder/bin/splunk _internal_launch_under_systemd
KillMode=mixed
KillSignal=SIGINT
TimeoutStopSec=360
LimitNOFILE = 1024000
LimitRTPRIO=99
SuccessExitStatus=51 52
RestartPreventExitStatus=51
RestartForceExitStatus=52
User=splunk
Group=splunk
NoNewPrivileges=yes
AmbientCapabilities=CAP_DAC_READ_SEARCH CAP_NET_ADMIN CAP_NET_RAW
ExecStartPre=-/bin/bash -c "chown -R splunk:splunk /opt/splunkforwarder"
Delegate=true
CPUShares=1024
MemoryLimit=4053409792
PermissionsStartOnly=true
ExecStartPost=-/bin/bash -c "chown -R splunk:splunk /sys/fs/cgroup/cpu/system.slice/%n"
ExecStartPost=-/bin/bash -c "chown -R splunk:splunk /sys/fs/cgroup/memory/system.slice/%n"
LimitNPROC = 512000
LimitFSIZE = infinity
LimitDATA = infinity
LimitCORE = infinity
TasksMax = infinity

[Install]
WantedBy=multi-user.target
```
I have not tested with CentOS/RHEL 9.
Make sure you are running a supported splunk version, The issue in that post refers to version 7.3.1, which has been out of support for almost a year. See [Splunk Products Version Compatibility Matrix](https://docs.splunk.com/Documentation/VersionCompatibility/current/Matrix/CompatMatrix#Splunk_Enterprise_version_7.x).

red667: Thanks for digging into this issue. I was using version 8.2.5. But I just merged with the new commits with the update to version 9.0.1. This splunk version creates the correct systemd file with cgroups.

So you can remove this merge request.

Thank you !!!

PR #130, opened 2022-08-23
fixes splunk/ansible-role-for-splunk#49

jewnix:
This allows you to have multiple apps in a single repo, and rsync the apps in the correct destination directory.

### DIFF
--- a/roles/splunk/tasks/install_apps.yml
+++ b/roles/splunk/tasks/install_apps.yml
@@ -90,7 +90,7 @@
 
 - name: Ensure correct permissions are set
   file:
-    path: "{{ splunk_home }}/{{ app_dest }}/{{ item.name }}"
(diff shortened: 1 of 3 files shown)


roles/splunk/tasks/install_apps.yml
```diff
@@ -90,7 +90,7 @@
 
 - name: Ensure correct permissions are set
   file:
-    path: "{{ splunk_home }}/{{ app_dest }}/{{ item.name }}"
+    path: "{{ splunk_home }}/{{ app_dest }}"
```
jewnix, line 93: If I did not remove this,Ansible will create a empty directory in `splunk_app_deploy_path` with the app name.
This will should have no impact, since the directory where the apps are should be owned by the user splunk is running as.

anelson-splunk (approved):

dtwersky (approved):

PR #133, opened 2022-09-20
Explicitly configure Splunk Group and User locally

arcsector:
# Summary

Explicitly configures the Splunk nix Group and User with the `local: true` stanza in order to ensure no clobbering of netgroups that may or may not be configured.

## Issues Fixed

When trying to add the group without the `local` config, if there is a netgroup configured with the same name as `splunk_nix_group` (say, for people who should be able to SSH into a system with LDAP creds), the group will not be created locally. This then causes the addition of the `splunk_nix_user` in the subsequent task to fail, as ansible will not add a local user to a netgroup.

Explicitly specifying `local: true` for each of these tasks solves this issue that some people may run into.

### DIFF
--- a/roles/splunk/tasks/install_splunk.yml
+++ b/roles/splunk/tasks/install_splunk.yml
@@ -6,6 +6,7 @@
       group:
         name: "{{ splunk_nix_group }}"
         state: present
+        local: true
       become: true
 
     - name: Add nix splunk user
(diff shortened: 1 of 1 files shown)


dtwersky (approved):

anelson-splunk (approved):

arcsector:
Thanks guys!

dtwersky:
This commit broke some distros that did not have `luseradd/lgroupadd` installed. I set the default to `false`, because that is how it was before this change.

Please update your group_vars to define `local_os_user` and `local_os_group` to true.

arcsector:
Probably a better idea, thanks.

Issue #48, opened 2021-03-05
Enhancement: Make downloading of packages and/or apps configurable to be local vs remote


mason-splunk:
This is somewhat related to https://github.com/splunk/ansible-role-for-splunk/issues/37

Need to add support for:
1. Downloading of Splunk packages on remote host
2. Downloading of apps/addons on remote host

Currently everything is downloaded to the local ansible host, and then rsync'd over. However, this should be configurable.

PR #134, opened 2022-09-30
feat(install): remove downloaded package after unarchive

zyphermonkey:
package is no longer needed after unarchive and can safely be removed to keep the disk clean


dtwersky:
@zyphermonkey The reason we decided to not remove the package is because if you run it against multiple instances, it will need to download the package many times. It annoys me sometime too to have to clean up old downloaded versions etc.

Do you have any other ideas how to clean it up without having to download it multiple times? Maybe a conditional var for that task when something like `cleanup: true` is passed at run-time? Maybe a tag?

zyphermonkey:
> @zyphermonkey The reason we decided to not remove the package is because if you run it against multiple instances, it will need to download the package many times. It annoys me sometime too to have to clean up old downloaded versions etc.

That makes sense. I forgot in my local version we don't use `delegate_to: localhost` anymore so the download and cleanup is always on a per host basis.

Maybe I'll working on getting another PR put in that allows for the user to chose local vs remote. Which would also close https://github.com/splunk/ansible-role-for-splunk/issues/48.
If download location is set to remote then we'll cleanup everytime. If not leave the download on the localhost for other hosts to prevent constant re-downloads.

zyphermonkey:
New commit adds the discussed feature although it doesn't completely close #48. I have not used the [configure_apps.yml](https://github.com/splunk/ansible-role-for-splunk/blob/master/roles/splunk/tasks/configure_apps.yml) download feature yet so I need some more time to review those tasks and determine the best way to implement local vs remote downloading.

Should be good to merge with just these changes now though. 😃

zyphermonkey:
Anything I need to do for merge?

dtwersky (approved):

PR #135, opened 2022-10-08
Add the code for the "Manager" option; have both old and new option r…

schneewe:
The README stated, that it is possible to have manager as value, but for me I didn't see any code for the Manager.
So I added the code for the manager and set the pass4SymmKey.

Additionally I changed the code, to remove the manager_uri, not even depending on the current splunk version.
No both the old and the new options were removed, in such cases were wee would change the license group, and splunk upgrade from version <9.0 to >= 9.x

### DIFF
--- a/roles/splunk/tasks/configure_license.yml
+++ b/roles/splunk/tasks/configure_license.yml
@@ -18,7 +18,7 @@
       group: "{{ splunk_nix_group }}"
     become: yes
     when:
-      - splunk_license_group=="Enterprise"
+      - splunk_license_group=="Enterprise" or splunk_license_group=="Manager"
   - name: Copy license file
     copy:
       src: "{{ item }}"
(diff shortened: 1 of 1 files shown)


dtwersky:
@schneewe That is an error in the `README.md`.  When the instance is a LM, you just set the `splunk_license_group` to `Enterprise`.

schneewe:
@dtwersky 
Yeah I see that only the Enterprise value is needed, but nothing in the role sets the pass4SymmKey in the general stanza for the LM. So it never activate the License Manager. That why I wrote the additional code and for me then it makes sense.
I also find a variable in the default vars "splunk_general_key" which is for the option, but again, nothing is using this.

dtwersky:
@schneewe you are correct about the `pass4SymmKey`. Strange that the variable does not exist anywhere.

Maybe we should add a var called `splunk_lm_key` just like there is `splunk_ds_key`, and this task should have something like:
`pass4SymmKey: "{{ splunk_lm_key }}"`, and set the same key for the LM and the Peers?

When splunk starts up, it will automatically set the general `pass4SymmKey` to `changeme`, so we should have something somewhere the option to set it, and be able to override it in this task. I'm also thinking that maybe we should set the general `pass4SymmKey` in the `configure_splunk_secret.yml` task when it's called by `install_splunk.yml`? 

I'm thinking if we should get rid of the `splunk_general_key` variable altogether, since it is not referenced anywhere, and replace it with a generic `pass4SymmKey`.

schneewe:
I will update the README and try to make a suggestion to solve this.
Honestly I would make the change in the configure_license.yml and check if the host is in the licensemaster group. Because not everyone is setting a global splunk_secret and so the task would not run. 
I use this role on many of my customers and never used the configure_secret task, because it's not necessary.

PR #136, opened 2022-10-08
add possibility to check a decrypted value, before setting a password…

schneewe:
I added a new task to check and validate a current encrypted value, so wee only set a password value only if they don't match. This results in lesser restarts, because the value is actually not changing. Otherwise in my current runs, the encrypted value gets overwritten, the task results in "changed" and splunk need a restart.

As an example I changed the workflow for the configure_deploymentclient.yml task.


schneewe:
Is this something to be helping or considering to implement?

dtwersky:
Just to make sure I understand this. This will prevent ansible from changing the encrypted `pass4SymmKey` every time it runs, or change it if it does not match the value of `splunk_ds_key`?

I like it being in a separate task, so it can be re-used for other encrypted values.

How have you tested this?

schneewe:
I already used it for one of my customers and at my company lab, because we wanted to optimize the workflow and reduce service restarts, due to no change of the pass4SymmKeys.
Actually it's pretty easy, the encrypted value gets decrypted with the splunk.secret and compared to the ansible variable.

I tested it with a running deployment, one time without any change, so nothing will happen and the playbook runs straight through to the end. if you change as an example the key for the deploymentserver pass4SymmKey then splunk restarts, as the file itself changed.

PR #137, opened 2022-10-11
Explicit Group Init in Boot Config

arcsector:
# Summary

Had to rebase a bit since my commits weren't attributed to my user, but the only commit needed is the [last one](2cf26163d98a3f5252d1b5e5878e188ba46f2499).  Essentially I'm just adding the `-group` flag to the boot start command since it's not there currently, and that causes some drift from the expected configs of the file permissions.

# Changes

Changed [roles/splunk/tasks/configure_splunk_boot.yml](https://github.com/splunk/ansible-role-for-splunk/commit/2cf26163d98a3f5252d1b5e5878e188ba46f2499) to include `-group {{ splunk_nix_group }}`


dtwersky:
Thanks for the PR. Sorry I was out for a bit, so I wasn't able to have a look at it.

This is a goo suggestion, but the reason it loops it because if a certain package does not exist, the whole operation will fail, and it will not install any packages at all.

This has been suggested before, but the packages should be split out by platform and version. For example, `policycoreutils-python` is RHEL only.

Any suggestions?

dtwersky:
Good suggestion. I tested this, and it seems that when you restart the splunk service, it gets set to the primary group the `splunk_nix_user` is set to. the `splunk_nix_group` is only set as an additional group to that user. The primary group is automatically set by the OS to the same of the user name when the user is created initially.

Some further research is needed to get this straightened out.

arcsector:
Awesome - thanks.

jewnix:
Now I'm thinking if [this](https://github.com/splunk/ansible-role-for-splunk/blob/master/roles/splunk/tasks/install_splunk.yml#L15) shoud have been `group` anstead of `groups`.

arcsector:
It definitely should be, but the issue with the splunk init boot config still requires a patch. I can add it to the PR if you'd like.

dtwersky (approved):

PR #138, opened 2022-10-12
feat(install_utilities): send list to package module

zyphermonkey:
Instead of looping through the list it is recommended to send a list to apt/yum when install multiple packages.

*When used with a `loop:` each package will be processed individually, it is much more efficient to pass the list directly to the name option.*   https://docs.ansible.com/ansible/latest/collections/ansible/builtin/yum_module.html

### DIFF
--- a/roles/splunk/defaults/main.yml
+++ b/roles/splunk/defaults/main.yml
@@ -72,19 +72,3 @@ add_pstack_script: false # Set to true to install a pstack generation script for
 configure_dmesg: false
 install_utilities: false # Set to true to install the list of packages defined in the linux_packages var after installing splunk
 use_tuned_thp: false
-linux_packages:
-  - nload
-  - iotop
-  - iftop
-  - sysstat
-  - telnet
-  - tcpdump
-  - htop
-  - atop
-  - lsof
-  - policycoreutils-python
-  - policycoreutils
-  - setroubleshoot
-  - nethogs
-  - gdb
-  - bind-utils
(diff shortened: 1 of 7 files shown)


dtwersky:
Thanks for the PR. Sorry I was out for a bit, so I wasn't able to have a look at it.

This is a goo suggestion, but the reason it loops it because if a certain package does not exist, the whole operation will fail, and it will not install any packages at all.

This has been suggested before, but the packages should be split out by platform and version. For example, policycoreutils-python is RHEL only.

Any suggestions?

zyphermonkey:
Okay to that explains the `ignore_errors: true` which I'm not a fan of and thought shouldn't be there.

I can come up with 2 idea.

1. remove the packages from the list that don't have common names between the various repositories.  
    This isn't ideal, but a quick fix
2. Move lists to [vars](https://github.com/splunk/ansible-role-for-splunk/blob/master/roles/splunk/vars/RedHat.yml)
    I like this idea the best, but will require me to move 
    ```yaml
    - name: Include OS family-specific variables
       include_vars: "{{ ansible_os_family }}.yml"
    ```
    from `configure_os.yml` to `main.yml` so the vars are available for all tasks.
    
    I'm willing to do the work, but want to make sure you're okay with the change first. 
    If so I'll need a list of package names for Debian as I only have access to RHEL currently.

zyphermonkey:
Thoughts?

zyphermonkey:
tested on `ubuntu2204` and `centos7`

dtwersky:
@zyphermonkey Thanks for the PR. At first glance, it looks good, and the includes being in `main.yml` makes a lot more sense.

I will try to test it on some more platforms and scenarios sometime this week.

dtwersky:
`RedHat8.yml` var file was added for RHEL8 specific packages.
Tested on Ubuntu 16/18/20/22.04 CentOS 7/8/8 Stream.

dtwersky (approved):

PR #143, opened 2022-11-08
forcing usage of luseradd/lgroupadd broke some distros

jewnix:
Some distros either don't have the `luseradd` and `lgroupadd` commands, or they are just broken.

This sets the `local: false` as the default (since this is how it has been before the recent change).
If you wish to force to create local user and group, set the `local_os_user` and `local_os_group` to `true`.


dtwersky (approved):

arcsector:
Good catch - super weird that ansible decides to use `luseradd` and doesn't have any fallback for systems which don't have that command, but that's an issue with ansible core not the splunk role. Thanks for this!

PR #144, opened 2022-11-16
Fixed least priv mode

jewnix:


roles/splunk/tasks/configure_facl.yml
```diff
@@ -41,6 +41,4 @@
         - restart non-redhat auditd service
       when: result_auditd_conf.stat.exists
 
-  when:
-    - least_privileged == false or "'full' in group_names"
-    - splunk_nix_user != 'root'
+  when: not least_privileged
```
jewnix, line 44: Moved over the conditional to `main.yml` so it can be re-used if needed.

roles/splunk/tasks/main.yml
```diff
@@ -51,7 +51,9 @@
     - name: Setting least priviledged mode
       set_fact:
         least_privileged: true
-      when: splunk_package_version is version(9.0, '>=')
+      when: >
```
jewnix, line 54: In version 9.0.0, a least privileged mode was introduced aromatically adding `CAP_DAC_READ_SEARCH` to the systemd unit file, which allows the forwarder to read all files on the filesystem, even it is running as a non-root user, therefore we don't need the `configure_facl.yml` task for those instances.

This feature is currently only available on the Universal Forwarder, therefore a `full` install will only be in `least_privileged` if it is running as root.

Issue #146, opened 2022-11-16
configure_thp.yml is skipped on RHEL 7+


lzannettacci:
the task is always skipped on RHEL 7+ machines because of the following conditions :
when:
    - (not ansible_os_family == "RedHat" and ansible_distribution_major_version|int > 6)
    - ansible_connection!="docker"
    - "'full' in group_names"

condition 1 seems to be the offending one

PR #147, opened 2022-11-16
changed thp conditional to ansible_service_manager Fixes splunk/ansible-role-for-splunk#146

jewnix:
The current conditional did not work right.
The limit on RedHat <6 was because of the lack of `systemd`. Changing to using `ansible_service_manager` will do the same thing, and is more descriptive of the reason why it's done that way.

I moved it to the `configure_os.yml` so it can skip the whole `include_tasks.yml` altogether instead of going through the whole task and skippin if it does not meet the requirements.


roles/splunk/tasks/configure_os.yml
```diff
@@ -12,7 +12,10 @@
 
 - name: Disable THP
   include_tasks: configure_thp.yml
-  when: "'full' in group_names"
+  when:
+    - ansible_service_mgr == "systemd"
```
anelson-splunk, line 16: Where is `ansible_service_mgr` defined?

roles/splunk/tasks/configure_os.yml
```diff
@@ -12,7 +12,10 @@
...
+    - ansible_connection != "docker"
+    - "'full' in group_names"
```
anelson-splunk, line 18: Are the double quotes around this necessary?

roles/splunk/tasks/configure_os.yml
dtwersky: It's a built-in fact

roles/splunk/tasks/configure_os.yml
dtwersky: Yes, it needs the double quotes

dtwersky (approved):

Issue #109, opened 2021-11-12
Logrotate config invalid


JasperJuergensen:
The logrotate config based on the `splunk_facl.j2` template created by the `configure_facl.yml` task seems to be invalid:

```
logrotate[685980]: error: splunk_facl:1 lines must begin with a keyword or a filename (possibly in double quotes)
logrotate[685980]: error: splunk_facl:5 unexpected }
logrotate[685980]: error: found error in file splunk_facl, skipping
```

Jalkar:
I believe, the file `/roles/splunk/templates/splunk_facl.j2` should be changed from : 
```{
    postrotate
        /usr/bin/setfacl -Rm u:{{ splunk_nix_user }}:rx /var/log
    endscript
}```

to 
```
/var/log {
    postrotate
        /usr/bin/setfacl -Rm u:{{ splunk_nix_user }}:rx /var/log
    endscript
}
```
Or something like that (but i don't have means to validate my assumption

nockstarr:
The problem has to do with ``\r`` characters. Just replace them at least that fixed the issue for me.
``tr -d '\r' < /etc/logrotate.d/splunk_facl> /etc/logrotate.d/splunk_facl``

lzannettacci:
@JasperJuergensen Had the same issue and applied what you suggest to solve it ! Thanks again

PR #152, opened 2022-11-22
Fixes splunk/ansible-role-for-splunk#109

jewnix:
Switch to using the `lineinfile` module instead of a template. This will add the `setfacl` command to where is belongs instead of in a separate script, which is prone to break.

### DIFF
--- a/roles/splunk/tasks/configure_facl.yml
+++ b/roles/splunk/tasks/configure_facl.yml
@@ -20,13 +20,12 @@
         - true
         - false
 
-    - name: Add logrotate script to enforce splunk user facls
-      template:
-        src: splunk_facl.j2
-        dest: /etc/logrotate.d/splunk_facl
-        owner: root
-        group: root
-      become: true
+    - name: Add setfacl to logrotate script
+      lineinfile:
+        path: /etc/logrotate.d/syslog
+        insertbefore: '  endscript'
+        line: '        /usr/bin/setfacl -Rm u:{{ splunk_nix_user }}:rx /var/log'
+      become: True
 
     - name: Check if auditd.conf is present
       stat:
--- a/roles/splunk/templates/splunk_facl.j2
+++ b/roles/splunk/templates/splunk_facl.j2
@@ -1,5 +1,0 @@
-{
-    postrotate
-        /usr/bin/setfacl -Rm u:{{ splunk_nix_user }}:rx /var/log
-    endscript
-}


midnight5ky (commented):
Just curious, will this insert the line specified each time this task is ran? Wasn't sure if it'll produce duplicates of the line since it just inserts the line before ' endscript'

jewnix:
> Just curious, will this insert the line specified each time this task is ran? Wasn't sure if it'll produce duplicates of the line since it just inserts the line before ' endscript'

In my testing, it has not put it in twice

midnight5ky (approved):

zyphermonkey:
@jewnix 
Would it be possible to revert this change?

1. I don't fully understand why this change was made.
   From the issue it seems like some were having issues with their line endings which was most likely cause by their .gitconfig settings and not the template. 
1. It's not best practice (nor am I a fan of) modifying default system config files. 
    Especially in the case of logrotate where it specifically gives us a `.d` directory to make our customization. 
1. Modifying default system config files can lead to inconsistent results.
    Because we have no control over how updates might change this `logrotate.d/syslog` file it's quite possible a system update will completely reset it destroying the changes made in this merge.
1. This merge isn't backwards compatible with previous version of the role.
    This doesn't cleanup the previous `logrotate.d/splunk_facl` file causing duplicate configs for existing systems.

jewnix:
@zyphermonkey  
> 1. I don't fully understand why this change was made.
>    From the issue it seems like some were having issues with their line endings which was most likely cause by their .gitconfig settings and not the template.

This was to fix #109 that failed because it did not have the log file directive in the script.
> 2. It's not best practice (nor am I a fan of) modifying default system config files.
>    Especially in the case of logrotate where it specifically gives us a `.d` directory to make our customization.

I usually refrain from modifying system files, however I did not find a way to solve this issue without this. Any recommendations would be appreciated.
> 3. Modifying default system config files can lead to inconsistent results.
>    Because we have no control over how updates might change this `logrotate.d/syslog` file it's quite possible a system update will completely reset it destroying the changes made in this merge.

I was not able to figure out a working solution. Adding `/var/log` in the old config file didn;t solve the issue, because no files were rotated, so the `postscript` did nothing.
> 4. This merge isn't backwards compatible with previous version of the role.
>    This doesn't cleanup the previous `logrotate.d/splunk_facl` file causing duplicate configs for existing systems.

Good point. Removing that file should be added

PR #155, opened 2022-11-30
Reset configure_boot_start var Fixes #150

### LINKED ISSUE #150: Playbook fails at ERROR: Misconfiguration detected! Unable to proceed as handlers will fail in the play later
My play is failing every time when i run multiple configuration files from a playbook. 
I have my variables `splunk_use_initd` and `splunk_use_sytemd` defined in group_vars/all.yml as follow:

```
splunk_use_initd: false  # If set to true, the system will use init.d. Default false
splunk_use_systemd: true # DO NOT EDIT. To use init.d, set `splunk_use_initd` to true.
```

My playbook contains:

```
-  hosts:
    - clustermanager
    - indexer
  roles:
    - ../roles/splunk
  vars:
    - deployment_task: check_splunk.yml

- hosts:
    - clustermanager
  roles:
    - ../roles/splunk
  vars:
    - deployment_task: configure_idxc_manager.yml

- hosts:
    - indexer
  roles:
    - ../roles/splunk
  vars:
    - deployment_task: configure_idxc_member.yml

```

This fails every time at configure_idxc_manager.yml task at the main.yml fail task:
`ERROR: Misconfiguration detected! Unable to proceed as handlers will fail in the play later."`

This is also confusing from roles/splunk/main.yml:

> Fail the play if the currently configured` boot-start method does match the expected state ` or boot-start is not enabled
> 
> **VS**
> 
> Either splunk boot-start is not enabled on this host, or its current `boot-start method does not matched the expected value` of splunk_use_initd/splunk_use_systemd

If set boot-start misconfigured var is being skipped but the `Fail the play if the currently configure boot-start...` still fails. 
```
TASK [../roles/splunk : Set boot-start misconfigured var] **********************
skipping: [splunk-host-indexer-manager]

```
Then it fails:
`ERROR: Misconfiguration detected! Unable to proceed as handlers will fail in the play later."`

This seems it's getting the value from the previous play (check_splunk.yml ) set_fact since these are the values being returned (even when `set boot-start misconfigured var is being` skipped):
```
ok: [splunk-host-indexer-manager] => {
     "msg": [
         true,
         true,
         "configure_idxc_manager.yml"
     ]
 }
```

Am I missing something?
Please let me know if any other information is needed.

Thanks in advance.

jewnix:
When running a play that installs splunk and then runs some additional tasks (like `splunk_idxc_deploy.yml`), it will fail at the second task because the `configure_boot_start` is still defined.

This resets the `configure_boot_start` to `False` at the start of the play, and fails only when the task sets it back to `True`

### DIFF
--- a/roles/splunk/tasks/main.yml
+++ b/roles/splunk/tasks/main.yml
@@ -7,10 +7,11 @@
     - "{{ ansible_os_family }}{{ ansible_distribution_major_version }}.yml"
     - "{{ ansible_os_family }}.yml"
 
-- name: Reset value of start_splunk_handler_fired
+- name: Reset value of start_splunk_handler_fired and configure_boot_start
   tags: always
   set_fact:
     start_splunk_handler_fired: false
+    configure_boot_start: False
   changed_when: false
 
 - name: Set boot-start method to systemd
@@ -131,7 +132,7 @@
       - "To correct this: Either run configure_splunk_boot.yml or update the value of splunk_use_initd/splunk_use_systemd in your group_vars."
   when:
     - splunkd_found.stat.exists
-    - configure_boot_start is defined
+    - configure_boot_start == True
     - not deployment_task == "configure_splunk_boot.yml"
 
 - name: Send Slack messages


arcsector:
Good catch on this one!

anelson-splunk (approved):

PR #156, opened 2022-11-30
Git app install changes

arcsector:
The Git apps install is nice, but it's SLOW. If you use this on a system that is even moderately loaded, you get huge performance hits. This PR is to help optimize the `install_apps` loop and to take out unnecessary components. The changes at a glance are:

- Moved cluster evaluation out of the `install_apps` loop
- Moved bundle push handler via notification out of the `install_apps` loop
- Created single file to call for bundle application with inferences
- Created task files for bundle application
- Call bundle application after loop in `configure_apps`

This really improves the loop functionality, as not only are bundle evals not being done inline, but this ports the bundle evals outside the loop to be reused by other plays or tasks. Having them as solely handlers is not good enough.


nockstarr:
Im interested in this PR. Maybe also add support for local apps at the same time? PR #81

arcsector:
Sorry this closed - I had to re-fork the repo

PR #157, opened 2022-11-30
Simple Firewall Configuration

arcsector:
This is an attempt to improve on https://github.com/splunk/ansible-role-for-splunk/pull/114 striving for simplicity while also maintaining future portability for people who want to add more ports or decrease the current port pool. The changes at a glance are:

- New port format to specify protocol for modularity
- Inferring of FW type: defaults to UFW and if RHEL uses Firewalld
- Added default firewall ports per group var
- Checks to see if firewall package is installed and service is running and enabled
- Using Firewalld Services instead of ephemeral commands
- Handlers for reloading firewalls and merged into single yaml file

Hopefully this is something that will be more useful. Comments appreciated - maybe we can move the init definition of `splunk_firewall_ports` to `group_vars/all.yml` but that's just an aesthetic choice I think.


roles/splunk/vars/RedHat.yml
```diff
@@ -1,19 +1,4 @@
 ---
 global_bashrc: /etc/bashrc
 chk_config_cmd: chkconfig --add disable-thp
-linux_packages:
```
dtwersky, line 4: Can you rebase this

roles/splunk/tasks/configure_os.yml
```diff
@@ -20,3 +20,9 @@
 - name: Enable read for dmesg
   include_tasks: configure_dmesg.yml
   when: configure_dmesg
+
+- name: Configure firewall service
+  include_tasks: "configure_firewall.yml"
+  when:
+    - firewall_service != 'undefined'
```
dtwersky, line 27: Why do we need this if we are already checking the `configure_firewall` var?

roles/splunk/vars/main.yml
```diff
@@ -6,3 +6,4 @@ splunk_build: "{{ splunk_package_url | regex_search('\\d+\\.\\d+\\.\\d+(?:\\.\\d
 # Create desired splunk version string (to compare with the output from the splunk version command for upgrades)
 splunk_version: "{{ splunk_product }} {{ splunk_v }} (build {{ splunk_build }})"
 splunk_auth: "{{ splunk_admin_username }}:{{ splunk_admin_password }}"
+firewall_service: ufw
```
dtwersky, line 9: Nit pick. no newline at the end of file.

roles/splunk/defaults/main.yml
```diff
@@ -31,6 +31,12 @@ splunk_admin_username: admin
 splunk_admin_password: undefined # Use ansible-vault encrypt_string, e.g. ansible-vault encrypt_string --ask-vault-pass 'var_value_to_encrypt' --name 'var_name'
 splunk_configure_secret: false # If set to true, you need to update files/splunk.secret
 splunk_secret_file: splunk.secret # Used to specify your splunk.secret filename(s), files should be placed in the "files" folder of the role
+configure_firewall: false # Whether or not to configure the firewall service on your machine, if set to true, opens firewall ports using UFW (default) or Firewalld depending on OS
+splunk_firewall_ports: # List of ports to allow through local firewall in dict form
```
dtwersky, line 35: I think it would be nice to have vars for the other ports as well so it can be customized, Something like `splunkweb_port` and `splunkhec_port` etc.

dtwersky:
@arcsector Thanks for creating this PR. I've wanted this for some time.

I was going through this, and I have a few thoughts. The group_vars for Indexers, Heavy Forwarders and search heads have a lot of things in common. How about we put the common ones in `full.yml`? That way we can avoid duplicate declarations.

In general, I wanted to add some vars that disable the web interface for IDX and HF as the default, as well as `splunktcp` for SH. I'm not sure if this will complicate things even more or not, but what are your thoughts?

Side note. Can you please for your repo directly from this one instead of the one from @zyphermonkey? This will prevent conflicts, and will be easier to be in sync with this upstream repo.

roles/splunk/tasks/configure_os.yml
arcsector: If someone unsets the firewall_service variable manually it should fail before getting to the playbook.

roles/splunk/vars/RedHat.yml
arcsector: My bad - thought I had done my cherry pick correctly and I did not.

roles/splunk/vars/main.yml
arcsector: Fixed in force push

roles/splunk/vars/RedHat.yml
arcsector: Fixed in force push

roles/splunk/defaults/main.yml
arcsector: Since I'm a novice at variable precedence, where should these be defined? I'm wishing they could be in their own separate file in `vars/Firewall.yml` but I don't know if that takes precedence over the `group_vars` - if it does, it wouldn't make it very useful, which leaves us the only option of defining them inline in `defaults/main.yml`. What do you think?

dtwersky: Here is a nice list of the precedences.
https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_variables.html#understanding-variable-precedence
I would put it in `defaults/main.yml`, and then uverride it per group in `group_vars` or per host in `host_vars`.

If you use `include_vars` in a task to include `Firewall.yml`, it'll override whatever you set in the var files.

roles/splunk/tasks/configure_firewall.yml
```diff
@@ -0,0 +1,33 @@
+---
+- name: Ensure {{ firewall_service }}  package is installed
+  ansible.builtin.package:
+    name: "{{ firewall_service }}"
+    state: present
```
dtwersky, line 5: You need to have `become: True`, otherwise the install will fail. You need this for all tasks that require superuser privileges.

roles/splunk/defaults/main.yml
arcsector: Alright for sure, it's just going to clog up the defaults file. As long as you're okay with that I'll proceed.

arcsector:
> I was going through this, and I have a few thoughts. The group_vars for Indexers, Heavy Forwarders and search heads have a lot of things in common. How about we put the common ones in `full.yml`? That way we can avoid duplicate declarations.

I have the common ones in `defaults/main.yml`, where the `splunk_firewall_ports` var is initialized - this will be applied to all other systems that have this in common (license, deploy, shdeployer, idx manager), then the only repeats are in the group_vars for the cluster components.

> In general, I wanted to add some vars that disable the web interface for IDX and HF as the default, as well as `splunktcp` for SH. I'm not sure if this will complicate things even more or not, but what are your thoughts?

I would like to do that, but I feel like that should be a separate PR which adds that config not just in the firewall config but also in `web.conf`. Also note I don't have `splunktcpin` opened for the SHC.

> Side note. Can you please for your repo directly from this one instead of the one from @zyphermonkey? This will prevent conflicts, and will be easier to be in sync with this upstream repo.

Will do.

roles/splunk/tasks/configure_firewall.yml
arcsector: Fixed with https://github.com/splunk/ansible-role-for-splunk/pull/157/commits/f93c8a18104136f20458ce214e3770068b9d1f80

roles/splunk/defaults/main.yml
arcsector: Done in https://github.com/splunk/ansible-role-for-splunk/pull/157/commits/f93c8a18104136f20458ce214e3770068b9d1f80

zyphermonkey:
Why aren't we just using the idempotent `firewall` module? Trying to manage XML templates doesn't add anything and just makes maintenance more complicated.

arcsector:
In my estimation Firewalld services are much more useful in terms of their overall introspection capabilities as well as their ease of modification and legibility. In addition, services are the best-practice in firewalld for things that are supposed to be associated with a given process. 

Say for example we're not just installing splunk on here, but also mysql. I can have a service file for splunk and then one for mysql which identify each port as being associated with that application, which i can then make modifications to or disable/enable independently of one another. A more targeted approach for extensible environments with a lot of firewall configs.

jewnix:
> Why aren't we just using the idempotent `firewall` module? Trying to manage XML templates doesn't add anything and just makes maintenance more complicated.

As @arcsector said. Although using the firewall plugin would may seem to be the simpler way, but when configuring multiple ports for a specific application, the recommended method from the OS side is services.

Also, running the `firewall-cmd` command is a lot slower, and running it multiple time every time slows down the playbook, which is slow enough as it is.

Any experienced sys-admin would know exactly where to look and what to look for.

roles/splunk/defaults/main.yml
```diff
@@ -75,3 +75,15 @@ add_pstack_script: false # Set to true to install a pstack generation script for
 configure_dmesg: false
 install_utilities: false # Set to true to install the list of packages defined in the linux_packages var after installing splunk
 use_tuned_thp: false
+# Firewall configs
+configure_firewall: false # Whether or not to configure the firewall service on your machine, if set to true, opens firewall ports using UFW (default) or Firewalld depending on OS
+# Firewall port presets - reference these in group_vars to assign them to splunk
+splunkweb_port: {protocol: "tcp", number: 8000}
+splunkhec_port: {protocol: "tcp", number: 8088}
+splunktcpin_port: {protocol: "tcp", number: 9997}
+splunkapi_port: {protocol: "tcp", number: "{{ splunkd_port }}"}
```
jewnix, line 84: I guess `splunkapi_port` is more descriptive of what it's actually for?

roles/splunk/tasks/configure_firewall.yml
```diff
@@ -0,0 +1,35 @@
...
+    src: firewalld_service.xml.j2
+    dest: /etc/firewalld/services/splunk.xml
+    backup: true
+    mode: 0644
+    owner: root
+    group: root
+  become: true
+  notify: reload firewalld
+  when: firewall_service == "firewalld"
+
```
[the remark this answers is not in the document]
jewnix: I'm going to commit a change to fix that. I'm running a few tests against it

roles/splunk/tasks/configure_firewall.yml
```diff
@@ -0,0 +1,35 @@
...
+- name: Configure UFW for Splunk
+  community.general.ufw:
```
nockstarr, line 28: What about adding a comment to UFW rules as well? Would also be more consistent since you are already using `<description>` in firewalld_service.xml.j2.

`comment: "{{ item.key }}"` should do the trick as the port var names is already fairly descriptive.

roles/splunk/tasks/configure_firewall.yml
```diff
@@ -0,0 +1,35 @@
+---
```
jewnix, line 1: I dont have permissions to push to your branch, but here is what I'm proposing.

A bit of an explanation on what is being done here:
ufw:
1. If `ufw` was not installed on the system, and the task is installing, it seems to not have the port 22 allowd, so we need to add that as well.
2. Enableing `ufw` with the service module, does not activate the firewall rules, so we are enabling it when we add port 22
3. When we add the splunk ports,, we trigger `reloaded` with the `ufw` module.

firewalld:
1. With `firewalld` we do need to enable and start the service, and port 22 is there by default.
2. After we add the service file, firewalld needs to be reloaded, otherwise we cannot enable the new service. We cannot use the handler to do it, because it will only run at the end of the play, so we need to do it manually before enabling the new service file
3. We need to enable the service for it to activate.

```
---
- name: Ensure {{ firewall_service }}  package is installed
  ansible.builtin.package:
    name: "{{ firewall_service }}"
    state: present
  become: true

- name: Configure firewalld for Splunk
  when: firewall_service == "firewalld"
  block:
    - name: Ensure {{ firewall_service }} is Started and Enabled
      ansible.builtin.systemd:
        name: "{{ firewall_service }}"
        state: started
        enabled: true
      become: true

    - name: Add splunk firewalld service
      ansible.builtin.template:
        src: firewalld_service.xml.j2
        dest: /etc/firewalld/services/splunk.xml
        backup: true
        mode: 0644
        owner: root
        group: root
      become: true
      register: firewalld

    - name: reload firewalld
      command: firewall-cmd --reload
      become: true
      when: firewalld.changed

    - name: Activate splunk firewalld service
      ansible.posix.firewalld:
        service: splunk
        permanent: yes
        state: enabled
        immediate: true
      notify: reload firewalld
      become: true

- name: Configure UFW for Splunk
  when: firewall_service == "ufw"
  block:
    - name: Ensure SSH is enabled
      community.general.ufw:
        port: "22"
        proto: "tcp"
        rule: "allow"
        state: "enabled"
      become: true

    - name: Add splunk port to UFW
      community.general.ufw:
        port: "{{ item.number }}"
        proto: "{{ item.protocol }}"
        rule: allow
        state: "reloaded"
      become: true
      loop: "{{ splunk_firewall_ports }}"
      when: firewall_service == "ufw"
```

jewnix (changes requested):

roles/splunk/vars/RedHat.yml
```diff
@@ -17,3 +17,4 @@ linux_packages:
   - nethogs
   - gdb
   - bind-utils
+firewall_service: firewalld
```
jewnix, line 20: You need to add this to the `RedHat8.yml` as well

roles/splunk/tasks/configure_firewall.yml
jewnix: You can also remove the `notify ufw` handler from `main.yml`, since it gets reloaded in the module itself.

roles/splunk/defaults/main.yml
arcsector: I think so, and it's distinct from the format to include as a dictionary

roles/splunk/tasks/configure_firewall.yml
arcsector: I added you as a contributor for future work, but I'll copypasta this for now, as well as add in @nockstarr s changes

roles/splunk/tasks/configure_firewall.yml
arcsector: Good suggestion - would love to have a XML comment in the Firewalld service that has the same concept - will this work if we're not iterating over key-value pairs though? Is `item.key` a reserved name when iterating over a list of variables? @jewnix any thoughts?

roles/splunk/vars/RedHat.yml
arcsector: Done in 1acc13b0aa4e7bb55809fb2ed801ffca86b3ba3f

roles/splunk/tasks/configure_firewall.yml
arcsector: Done in 1acc13b0aa4e7bb55809fb2ed801ffca86b3ba3f

roles/splunk/tasks/configure_firewall.yml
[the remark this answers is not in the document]
arcsector: Done in 1acc13b0aa4e7bb55809fb2ed801ffca86b3ba3f

roles/splunk/tasks/configure_firewall.yml
nockstarr: @arcsector you are right we cant use `item.key` in a dictionary so we would have to define a `desc` key in the dictionary.
Ex `splunkweb_port: { protocol: "tcp", number: 8000, desc: "Splunk Web" }` 

We can implement `comment: ` as optional like this as UFW accepts an empty string as comment.

```yaml
- name: Add splunk port to UFW
      community.general.ufw:
        port: "{{ item.number }}"
        proto: "{{ item.protocol }}"
        comment: "{{ item.desc | default('') }}"
        rule: allow
        state: reloaded
      become: true
      loop: "{{ splunk_firewall_ports }}"

# TASK [Add splunk port to UFW] **************************************************************************************
# changed: [localhost] => (item={'protocol': 'tcp', 'number': 8000, 'desc': 'Splunk Web'})
# changed: [localhost] => (item={'protocol': 'tcp', 'number': 8089})
#
#
# $ sudo ufw status verbose
#
# To                         Action      From
# --                         ------      ----
# 8000/tcp                   ALLOW IN    Anywhere                   # Splunk Web
# 8089/tcp                   ALLOW IN    Anywhere   
```

arcsector: This would definitely be good if you had a non-standard value for ports, but I don't know if we want to complicate the dictionary more. So i do like the idea of making the default value for the description, so that sounds like a plan. I'll push that too.

jewnix: I'm all for adding descriptions.

roles/splunk/defaults/main.yml
```diff
@@ -75,3 +75,16 @@ add_pstack_script: false # Set to true to install a pstack generation script for
...
+configure_firewall: false # Whether or not to configure the firewall service on your machine, if set to true, opens firewall ports using UFW (default) or Firewalld depending on OS
+splunk_firewall_service: splunk # The name of the Splunk firewall service to install for firewalld
+# Firewall port presets - reference these in group_vars to assign them to splunk
+splunkweb_port: {protocol: "tcp", number: 8000}
+splunkhec_port: {protocol: "tcp", number: 8088}
+splunktcpin_port: {protocol: "tcp", number: 9997}
+splunkapi_port: {protocol: "tcp", number: "{{ splunkd_port }}"}
+splunkidxcrep_port: {protocol: "tcp", number: "{{ splunk_idxc_rep_port }}"}
+splunkshcrep_port: {protocol: "tcp", number: "{{ splunk_shc_rep_port }}"}
+splunk_firewall_ports: # List of ports to allow through local firewall in dict form
```
jewnix, line 88: Do we really need to define this? This will open the ports on UFs, and since the UF is no longer listening on 0.0.0.0:8089 by default, I don't think we need that.

roles/splunk/tasks/configure_firewall.yml
```diff
@@ -0,0 +1,61 @@
...
+- name: Ensure {{ firewall_service }} package is installed
+  ansible.builtin.package:
+    name: "{{ firewall_service }}"
+    state: present
+  become: true
+
+- name: Configure firewalld for Splunk
+  block:
+    - name: Ensure firewalld is Started and Enabled
```
jewnix, line 10: I changed this to say `firewalld` instead of `firewall_service` because it was a bit confusing in the output when it ran on Debian. It showed `skipped` and name said `ufw`.

roles/splunk/defaults/main.yml
arcsector: Thankfully it will not - in [`configure_os.yml`](https://github.com/splunk/ansible-role-for-splunk/pull/157/commits/f93c8a18104136f20458ce214e3770068b9d1f80#diff-c80bf73fffa3043683ec15ad3dc6d7b4e303e31a2aacd745625d7e1b09218f96) I have this so that we don't install the firewall unless it's in the `full` category:

```yaml
- name: Configure firewall service
  include_tasks: "configure_firewall.yml"
  when:
    - firewall_service != 'undefined'
    - configure_firewall != false
    - "'full' in group_names"
```

roles/splunk/tasks/configure_firewall.yml
arcsector: Done in https://github.com/splunk/ansible-role-for-splunk/pull/157/commits/4c7b5f3381933bc6473b3c48df1ee23cda16ef8e

jewnix (approved):

PR #158, opened 2022-12-06
Fix setfacl task

masonsmorales:
A [recent commit](https://github.com/splunk/ansible-role-for-splunk/commit/9df2892e86753793aa4e1d653ebd58c36b47df3e) changed the module that was used in `configure_facl.yaml` and causes the role to fail with a `file not found` error during splunk installation when `path: /etc/logrotate.d/syslog` does not already exist on the target host. 

This is due to the [lineinfile module](https://docs.ansible.com/ansible/latest/collections/ansible/builtin/lineinfile_module.html#parameters) defaulting the `create` parameter to `false`. This PR sets `create: true` to remedy the issue.

### DIFF
--- a/roles/splunk/tasks/configure_facl.yml
+++ b/roles/splunk/tasks/configure_facl.yml
@@ -25,7 +25,8 @@
         path: /etc/logrotate.d/syslog
         insertbefore: '  endscript'
         line: '        /usr/bin/setfacl -Rm u:{{ splunk_nix_user }}:rx /var/log'
-      become: True
+        create: true
+      become: true
 
     - name: Check if auditd.conf is present
       stat:


jewnix:
Some distro (debain based) have the filename as rsyslog instead of syslog.
Creating the file may end up not working since it will not put anything in there

masonsmorales:
> Some distro (debain based) have the filename as rsyslog instead of syslog. Creating the file may end up not working since it will not put anything in there

Sounds like there should be a distro-specific var there instead. I'll close this PR.

jewnix:
@masonmorales [This](https://github.com/splunk/ansible-role-for-splunk/pull/160) PR fixes this issue as well.
Can you give it a try? Maybe I overlooked some versions.

PR #160, opened 2022-12-11
Require acl package. Fix logrotate and auditd

jewnix:
The `Check for existing /opt/splunkforwarder/etc/passwd` fails if the `acl` package is not installed. Observed on Debian servers with minimal installation.

The `prereqs.yml` can be used in the future to install packages required for the basic installation operation, even if `install_utilities` is set to false.

### DIFF
--- a/roles/splunk/handlers/main.yml
+++ b/roles/splunk/handlers/main.yml
@@ -73,7 +73,9 @@
   when: not start_splunk_handler_fired
 
 - name: restart redhat auditd service
-  command: service auditd condrestart
+  shell: |
+    service auditd stop
+    service auditd start
   become: true
   when: ansible_os_family == 'RedHat'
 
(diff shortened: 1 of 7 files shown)


roles/splunk/handlers/main.yml
```diff
@@ -73,7 +73,9 @@
   when: not start_splunk_handler_fired
 
 - name: restart redhat auditd service
-  command: service auditd condrestart
+  shell: |
+    service auditd stop
```
guohuang, line 77: why do you need to stop and start the service here?

jewnix: Yes, That's according to the RedHat documentation
https://access.redhat.com/solutions/2664811

guohuang (approved):

PR #161, opened 2022-12-14
Remove Splunk additions

arcsector:
Adding some removals to `remove_splunk.yml` for better cleanup:

- User and Group
- THP/Ulimit files & services
- Dmesg files
- Firewall files and entries
- FACL files and lineinfile's

### DIFF
--- a/roles/splunk/tasks/remove_splunk.yml
+++ b/roles/splunk/tasks/remove_splunk.yml
@@ -31,17 +31,6 @@
 
 - name: Remove Disable-THP/Ulimit files
   block:
-    - name: Stop Disable-THP service
-      ansible.builtin.service:
-        name: disable-thp
-        enabled: false
-        state: stopped
-      become: true
-      when:
-        - ansible_connection!="docker"
-        - "'full' in group_names"
-      ignore_errors: true
-
     - name: Remove disable-thp service to systemd (CentOS/RHEL 7+)
       ansible.builtin.file:
         path: /etc/systemd/system/disable-thp.service
@@ -50,8 +39,9 @@
       notify:
         - reload systemctl daemon
       when:
-        - ansible_os_family == "RedHat"
-        - ansible_distribution_major_version|int > 6
+        - not use_tuned_thp
+        - ansible_connection != "docker"
+        - "'full' in group_names"
 
     - name: Remove ulimits
       ansible.builtin.file:
--- a/roles/splunk/tasks/remove_splunk.yml
+++ b/roles/splunk/tasks/remove_splunk.yml
@@ -13,7 +13,7 @@
 
 - name: Block for non-root splunk user removal
   block:
-    - name: Add nix splunk user
+    - name: Remove nix splunk user
       ansible.builtin.user:
         name: "{{ splunk_nix_user }}"
         groups: "{{ splunk_nix_group }}"
@@ -62,7 +62,7 @@
 
 - name: Remove Dmesg configs
   block:
-    - name: Add sysctl for dmesg
+    - name: Remove sysctl for dmesg
       ansible.posix.sysctl:
         name: kernel.dmesg_restrict
         value: '0'
@@ -118,7 +118,7 @@
 
 - name: Remove FACLs
   block:
-    - name: Add setfacl to logrotate script
+    - name: Remove setfacl to logrotate script
       ansible.builtin.lineinfile:
         path: /etc/logrotate.d/syslog
         insertbefore: '  endscript'
@@ -132,7 +132,7 @@
       register: result_auditd_conf
       become: true
 
-    - name: Allow splunk to read /var/log/audit.log
+    - name: Remove line from /var/log/audit.log
       community.general.ini_file:
         path: /etc/audit/auditd.conf
         section: null
--- a/roles/splunk/tasks/remove_splunk.yml
+++ b/roles/splunk/tasks/remove_splunk.yml
@@ -102,8 +102,7 @@
   block:
     - name: Remove setfacl to logrotate script
       ansible.builtin.lineinfile:
-        path: /etc/logrotate.d/syslog
-        insertbefore: '  endscript'
+        path: "{{ logrotate_file }}"
         line: '        /usr/bin/setfacl -Rm u:{{ splunk_nix_user }}:rx /var/log'
         state: absent
       become: true
--- a/roles/splunk/tasks/remove_splunk.yml
+++ b/roles/splunk/tasks/remove_splunk.yml
@@ -118,8 +118,8 @@
         path: /etc/audit/auditd.conf
         section: null
         option: log_group
-        value: "{{ splunk_nix_group }}"
-        state: absent
+        value: root
+        state: present
       become: true
       notify:
         - restart redhat auditd service
--- a/roles/splunk/tasks/remove_splunk.yml
+++ b/roles/splunk/tasks/remove_splunk.yml
@@ -60,25 +60,7 @@
       become: true
       when: splunk_use_initd
 
-- name: Remove Dmesg configs
-  block:
-    - name: Remove sysctl for dmesg
-      ansible.posix.sysctl:
-        name: kernel.dmesg_restrict
-        value: '0'
-        state: absent
-        sysctl_file: /etc/sysctl.d/20-splunk.conf
-        reload: true
-      register: sysctl_dmesg
-      become: true
-
-    - name: Restart procps
-      ansible.builtin.service:
-        name: procps
-        state: restarted
-      when: sysctl_dmesg.changed
-      become: true
-  when: configure_dmesg
+# Not removing DMesg ini config due to defaults on some systems breaking
 
 - name: Remove firewall artifacts
   block:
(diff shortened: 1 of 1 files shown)


arcsector:
Wanted to note that I was going to include the removal of the polkits from 41b2a8e98c597d0a49b238ccbf1713edd141fa46, however we have to reload dbus to do that, and in my tests it's very precarious reloading dbus, sometimes resulting in a kernel panic from crashing dbus, and so I left it out until someone more experienced with that can take that on.

roles/splunk/tasks/remove_splunk.yml
```diff
@@ -11,4 +11,139 @@
...
+
+- name: Remove FACLs
+  block:
+    - name: Add setfacl to logrotate script
+      ansible.builtin.lineinfile:
+        path: /etc/logrotate.d/syslog
+        insertbefore: '  endscript'
+        line: '        /usr/bin/setfacl -Rm u:{{ splunk_nix_user }}:rx /var/log'
+        state: absent
+      become: true
```
nockstarr, line 127: NB Likely gonna be updated in https://github.com/splunk/ansible-role-for-splunk/pull/160/files#diff-032c335b5fa8c728153c887ba6051c71b11474b8ba98682616b9ac79d2605968

roles/splunk/tasks/remove_splunk.yml
```diff
@@ -11,4 +11,139 @@
...
+    - name: Remove ulimits
+      ansible.builtin.file:
+        path: /etc/security/limits.d/splunk_ulimits.conf
+        state: absent
+      become: true
+      when: splunk_use_initd
+
+- name: Remove Dmesg configs
+  block:
+    - name: Add sysctl for dmesg
```
nockstarr, line 65: Remove?

roles/splunk/tasks/remove_splunk.yml
```diff
@@ -11,4 +11,139 @@
     state: absent
   become: true
 
-# Todo: Remove user, group, THP/ulimit files, check for cron jobs/scripts and remove those if present
+- name: Block for non-root splunk user removal
+  block:
+    - name: Add nix splunk user
```
nockstarr, line 16: Remove?

roles/splunk/tasks/remove_splunk.yml
arcsector: Let me know when it's merged - definitely better to have it as a var, that should be a PR all of it's own, to do:

- Ulimit file
- THP Service file
- DMesg file
- Auditd conf file

roles/splunk/tasks/remove_splunk.yml
arcsector: Fixed, along with a few other names, in https://github.com/splunk/ansible-role-for-splunk/pull/161/commits/b70c6e5be6e4238a9968acf0d00512ef849ca656

roles/splunk/tasks/remove_splunk.yml
arcsector: Fixed, along with a few other names, in https://github.com/splunk/ansible-role-for-splunk/pull/161/commits/b70c6e5be6e4238a9968acf0d00512ef849ca656

jewnix: Don't remove the sysctl. Some systems have it this way by default, and we don't want to modify it.

roles/splunk/tasks/remove_splunk.yml
```diff
@@ -11,4 +11,139 @@
...
+      loop: "{{ splunk_firewall_ports }}"
+      when: firewall_service == "ufw"
+  when:
+    - firewall_service != 'undefined'
+    - configure_firewall != false
+    - "'full' in group_names"
+
+- name: Remove FACLs
+  block:
+    - name: Remove setfacl to logrotate script
```
jewnix, line 121: I think I should get [this](https://github.com/splunk/ansible-role-for-splunk/pull/160) PR merged. The logrotate is broken for some systems.

jewnix:
I actually have a local branch to remove the user and group for a while :smile: , I just didn't push it

roles/splunk/tasks/remove_splunk.yml
arcsector: Interesting - definitely let me know when you have that.

roles/splunk/tasks/remove_splunk.yml
arcsector: Removed dmesg removal in https://github.com/splunk/ansible-role-for-splunk/pull/161/commits/fd19974459c7332a846d5653cb4742a6b0cd59ef

dtwersky:
> Wanted to note that I was going to include the removal of the polkits from [41b2a8e](https://github.com/splunk/ansible-role-for-splunk/commit/41b2a8e98c597d0a49b238ccbf1713edd141fa46), however we have to reload dbus to do that, and in my tests it's very precarious reloading dbus, sometimes resulting in a kernel panic from crashing dbus, and so I left it out until someone more experienced with that can take that on.

when you run `disable boot-start` it should remove the polkit as well. At least that's the expected behavior.

roles/splunk/tasks/remove_splunk.yml
jewnix: I merged PR. You can adjust this accordingly

roles/splunk/tasks/remove_splunk.yml
```diff
@@ -11,4 +11,121 @@
...
+
+    - name: Remove nix splunk group
+      ansible.builtin.group:
+        name: "{{ splunk_nix_group }}"
+        state: absent
+        local: "{{ local_os_group }}"
+      become: true
+  when: splunk_nix_user != 'root'
+
+- name: Remove Disable-THP/Ulimit files
```
jewnix, line 32: There is also a `tuned` profile to disable THP, but I think you can leave that in there for now, because we don't really know what profile it was using prior to the change. Unless we record somewhere what he original profile was. But that's a whole other thing that I wanted to do (adding a comment in the new file what the original was), and is probably outside the scope of this PR.

roles/splunk/tasks/remove_splunk.yml
arcsector: Done in https://github.com/splunk/ansible-role-for-splunk/pull/161/commits/9506074507a20fe35e470476465553c013b1a8f2 - great PR, good catch on the auditd restart.

roles/splunk/tasks/remove_splunk.yml
arcsector: Agreed, the tuned profile switch is a great option for people who aren't already using tuned, and I'm sure they can change it accordingly in their removal file.

arcsector:
@jewnix in some of my tests on RHEL `procps` doesn't exist as a service - is this the same for you? Should we add a conditional for debian machines only?

roles/splunk/tasks/remove_splunk.yml
```diff
@@ -11,4 +11,120 @@
...
+        state: absent
+      become: true
+
+    - name: Check if auditd.conf is present
+      ansible.builtin.stat:
+        path: /etc/audit/auditd.conf
+      register: result_auditd_conf
+      become: true
+
+    - name: Remove line from /var/log/audit.log
```
dtwersky, line 116: This may cause a problem because it's removing that line from the conf file. I'm not sure how auditd will behave when that line is not there. The value needs to be set to `root`.

arcsector: That works for me.

arcsector:
> it would be a good idea to add a conditional for `ansible_os_family` on at least the restart of this service, if not on the whole task.

Gotcha. Separate PR for sure.

roles/splunk/tasks/remove_splunk.yml
arcsector: Fixed in https://github.com/splunk/ansible-role-for-splunk/pull/161/commits/cf45da572ce0537de3d6eb6731700f004a78c51f

roles/splunk/tasks/remove_splunk.yml
```diff
@@ -11,4 +11,120 @@
...
+      ansible.builtin.group:
+        name: "{{ splunk_nix_group }}"
+        state: absent
+        local: "{{ local_os_group }}"
+      become: true
+  when: splunk_nix_user != 'root'
+
+- name: Remove Disable-THP/Ulimit files
+  block:
+    - name: Stop Disable-THP service
```
dtwersky, line 34: Sorry for the nit-pick, but this service does not need to be stopped, because the exec exits right after it runs, so removing the unit file and reloading systemd should be enough.
Also, the conditional for removing the unit file should be:
```
      when:
        - not use_tuned_thp
        - ansible_connection!="docker"
        - "'full' in group_names"
```

arcsector: Done - fixed in [d31f002](https://github.com/splunk/ansible-role-for-splunk/pull/161/commits/d31f00288c5dbbb974b508df1cb30175f8c5443d)

dtwersky (approved):

PR #172, opened 2023-02-06
fix(configure_license): use version_compare

zyphermonkey:
comparing float doesn't work with MAJOR.MINOR.PATCH semantic versioning Splunk uses.

### DIFF
--- a/roles/splunk/tasks/configure_license.yml
+++ b/roles/splunk/tasks/configure_license.yml
@@ -5,7 +5,7 @@
 
 - name: Setting licensing mode based on Splunk version number
   set_fact:
-    mode_option: "{{ splunk_version_release is version_compare('9.0', '<') | ternary('master_uri', 'manager_uri') }}"
+    mode_option: "{{ splunk_version_release is version('9.0', '<') | ternary('master_uri', 'manager_uri') }}"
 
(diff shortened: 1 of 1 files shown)


roles/splunk/tasks/configure_license.yml
```diff
@@ -5,7 +5,7 @@
 
 - name: Setting licensing mode based on Splunk version number
   set_fact:
-    mode_option: "{% if splunk_version_release | float < 9.0 %}master{% else %}manager{% endif %}_uri"
+    mode_option: "{{ splunk_version_release is version_compare('9.0', '<') | ternary('master_uri', 'manager_uri') }}"
```
jewnix, line 8: Good catch. `version_compare` is now `version`. I am using it [here](https://github.com/splunk/ansible-role-for-splunk/blob/master/roles/splunk/tasks/main.yml#L67).

dtwersky: I tested, seems to be working fine.

Just a nit-pick, can you just change `version_compare` to `version`? It was renamed in version 2.5 [reference](https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_tests.html)

PR #173, opened 2023-02-09
Splunk license file

zyphermonkey:
Allow license files to be stored in custom src dir for better organization.

Retains backwards compatability by using `default` filter.

Example Var definition
```yaml
splunk_license_file:
  - src: licenses/splunk1.lic
    dest: splunk1.lic
  - src: licenses/Splunk2.lic
    dest: Splunk2.lic
```

### DIFF
--- a/roles/splunk/tasks/configure_license.yml
+++ b/roles/splunk/tasks/configure_license.yml
@@ -21,8 +21,8 @@
       - splunk_license_group=="Enterprise"
   - name: Copy license file
     copy:
-      src: "{{ item }}"
-      dest: "{{ splunk_home }}/etc/licenses/enterprise/{{ item }}"
+      src: "{{ item.src | default(item) }}"
+      dest: "{{ splunk_home }}/etc/licenses/enterprise/{{ item.dest | default(item) }}"
       owner: "{{ splunk_nix_user }}"
       group: "{{ splunk_nix_group }}"
       mode: "0600"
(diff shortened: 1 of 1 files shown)


roles/splunk/tasks/configure_license.yml
```diff
@@ -50,6 +51,18 @@
...
+      path: "{{ splunk_home }}/etc/system/local/server.conf"
+      section: general
+      option: pass4SymmKey
+      value: "{{ splunk_general_key }}"
+      owner: "{{ splunk_nix_user }}"
+      group: "{{ splunk_nix_group }}"
+    become: yes
+    notify: restart splunk
+    when:
+      - splunk_license_group=="Enterprise"
```
jewnix, line 65: This should be `when "'licensemaster' in group_names"`.

It does not seem that much thought was put in to the LM configuration. I think there should be a `splunk_lm_key` variable just like there is one for the idxc etc. even though this one is in the general stanza. The default value in `main.yml` can be `splunk_lm_key: "{{ splunk_general_key }}"`, but it gives you the option to customize it based on the role.

If done this way, the pass4SymmKey value for the license peer, should also be changed to `splunk_lm_key`.

schneewe: I'm not sure if it is a good way to only check `when "'licensemaster' in group_names"`, because what if someone has an additional standalone server with it's own license. I think the naming licensemaster can than be misleading.

I do not agree with a new variable, as the licensing does not have an own pass4SymmKey. It uses the one under the general stanza. So changing the pass4SymmKey besides the `splunk_general_key` generates different configurations.

PR #174, opened 2023-02-09
feat(cron): set permissions on cron scripts/jobs to splunk

zyphermonkey:
Following the "least privilege" model we should only be running things as root when absolutely necessary.

Also migrate from crontab to cron.d to isolate jobs into separate files for more granular management.

### DIFF
--- a/roles/splunk/tasks/add_crashlog_script.yml
+++ b/roles/splunk/tasks/add_crashlog_script.yml
@@ -3,8 +3,8 @@
   template:
     src: cleanup_crashlogs.sh.j2
     dest: "{{ splunk_home }}/cleanup_crashlogs.sh"
-    owner: root
-    group: root
+    owner: "{{ splunk_nix_user }}"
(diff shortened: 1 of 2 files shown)


zyphermonkey:
Since this changes the location of the cron configuration do we want to add cleanup tasks for the crontab entries?

I couldn't find any other references for cleanup for any other changes so I assume the idea is that this just needs to work on a fresh system.

I can add the cleanup tasks though if requested, just being cautious of adding tasks that will go unused almost all the time.

jewnix:
@zyphermonkey 
> If we keep the old cron job which runs as root, and change the owner to splunk_nix_user, that will allow the splunk user to modify the script, and let it do nasty stuff as the root user. This is a lot more dangerous than leaving it the way it is now. 

On second thought, we can change the mode to `0555` so the splunk user cannot modify that file. But I think there should be something there to at least attempt to clean up the old cron job.

PR #177, opened 2023-03-21
KVStore Tools

arcsector:
# Summary

This PR provides additional KVStore tools available to the user to be configured, including:

- Disable
- Backup
- Oplog and storage engine set at install time
- KVStore Storage Engine & Server Version Upgrade


arcsector:
Possible other features to include:

- KVStore Restore from backup
- KVStore Clean
- KVStore Resync
- Oplog change in SHC

roles/splunk/tasks/adhoc_backup_kvstore.yml
```diff
@@ -0,0 +1,15 @@
...
+    {{ splunk_home }}/bin/splunk backup kvstore {{ archive_name | default("") }}
+  become: true
+  become_user: "{{ splunk_nix_user }}"
+  register: splunk_kvstore_backup_out
+  changed_when: splunk_kvstore_backup_out.rc == 0
+  failed_when: splunk_kvstore_backup_out.rc != 0
+
+- name: Check that backup has finished
+  ansible.builtin.command: |
+    {{ splunk_home }}/bin/splunk splunk show kvstore-status | grep backupRestoreStatus | sed -r 's/\s+backupRestoreStatus : //g'
```
dtwersky, line 13: This need authentication too

roles/splunk/tasks/kvstore_upgrade.yml
```diff
@@ -0,0 +1,152 @@
+---
+- name: Set fact for kvstore Upgrade
+  ansible.builtin.set_fact:
+    splunk_upgrade_server: false
+
+- name: Check the current kvstore storage backend
+  ansible.builtin.command: |
+    {{ splunk_home }}/bin/splunk show kvstore-status --verbose | grep storageEngine | sed -r 's/\s+storageEngine : //g'
```
dtwersky, line 8: This need authentication too

roles/splunk/tasks/kvstore_upgrade.yml
```diff
@@ -0,0 +1,152 @@
...
+  become: true
+  become_user: "{{ splunk_nix_user }}"
+  register: splunk_kvstore_backend_out
+  changed_when: splunk_kvstore_backend_out.rc == 0
+  failed_when: splunk_kvstore_backend_out.rc != 0
+
+- name: Check the current kvstore server version
+  ansible.builtin.command: |
+    {{ splunk_home }}/bin/splunk show kvstore-status --verbose | grep serverVersion | sed -r 's/\s+serverVersion : //g'
```
dtwersky, line 17: Well, I'm not going to repeat myself on the authentication :)

roles/splunk/tasks/kvstore_upgrade.yml
dtwersky, line 8: Works without `--verbose` too

dtwersky:
@arcsector This is a great PR, especially the migration part. As I mentioned in the other comments, the commands need authentication, so we either need to add them, or start the whole task with `splunk login` so we don't need to set `no_log` on all the commands. I have not tested the `splunk login` method, but I assume it would work.

roles/splunk/tasks/kvstore_upgrade.yml
arcsector: Docs say to use `--verbose` so I'm going to keep it that way: https://docs.splunk.com/Documentation/Splunk/latest/Admin/MigrateKVstore

arcsector:
Fixed all the auth issues (sorry it slipped my mind) in https://github.com/splunk/ansible-role-for-splunk/pull/177/commits/ce2c80a4bbd06fedb016c01c403d2b927a4abdf6

roles/splunk/tasks/configure_kvstore.yml
```diff
@@ -0,0 +1,28 @@
+---
+- name: Disable KVStore if specified
```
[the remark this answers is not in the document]
arcsector: Fixed in https://github.com/splunk/ansible-role-for-splunk/pull/177/commits/ce2c80a4bbd06fedb016c01c403d2b927a4abdf6

roles/splunk/tasks/adhoc_clean_kvstore.yml
```diff
@@ -0,0 +1,14 @@
+---
+- name: Stop Splunkd service
+  include_tasks: splunk_stop.yml
+
+- name: Clean KVStore
+  ansible.builtin.command: "{{ splunk_home }}/bin/splunk -auth {{ splunk_auth }} clean kvstore --local --answer-yes"
```
jewnix, line 6: Cleaning the KV Store does not require authentication, since splunk is stopped. If you don't mind, you can add me as a collaborator to your fork, and I can add a task to login to splunk once, so we can avoid all this login junk, which has been getting on my last nerve. Requiring `no_log` inclusion for these tasks makes it difficult to troubleshoot when things don't work.

arcsector: Done, I just added you. I would be very interested in a Splunk Login task in a separate file so that it can be called as needed...

jewnix: Pushed. I have not tested it yet, so I may have overlooked something. But you can review and test it if you want to, and let me know if I missed something.

I can test it, but I'll need to spin up a few environments to do so, and it can take some time.

roles/splunk/tasks/kvstore_upgrade.yml
```diff
@@ -0,0 +1,156 @@
...
+  ansible.builtin.debug:
+    var: splunk_current_server_version_out.stdout
+    verbosity: 2
+
+- name: Upgrade KVstore if needed
+  block:
+    - name: Backup KVStore
+      include_tasks: adhoc_backup_kvstore.yml
+      vars:
+        - archive_name: "-archiveName preAnsibleVersionUpgradeBackup"
```
jewnix, line 43: Should this maybe be customizable?

roles/splunk/tasks/kvstore_upgrade.yml
```diff
@@ -0,0 +1,156 @@
+---
+- name: Check if authenticated
+  include_tasks: splunk_login.yml
+  when: not splunk_authenticated
+
+- name: Set fact for kvstore Upgrade
+  ansible.builtin.set_fact:
+    splunk_upgrade_server: false
```
jewnix, line 8: Where is this var being used?

roles/splunk/tasks/kvstore_upgrade.yml
```diff
@@ -0,0 +1,156 @@
...
+
+- name: Check the current kvstore storage backend
+  ansible.builtin.command: |
+    {{ splunk_home }}/bin/splunk show kvstore-status | grep storageEngine | sed -r 's/\s+storageEngine : //g'
+  become: true
+  become_user: "{{ splunk_nix_user }}"
+  register: splunk_kvstore_backend_out
+  changed_when: splunk_kvstore_backend_out.rc == 0
```
jewnix, line 16: I personally like to set `changed_when: false` for checks. In my head it makes sense.

roles/splunk/tasks/kvstore_upgrade.yml
arcsector: Eh, not in my opinion, but it's trivial to do so.

roles/splunk/tasks/kvstore_upgrade.yml
arcsector: Oof, that's a headpalm

roles/splunk/tasks/kvstore_upgrade.yml
arcsector: fair play

arcsector: Fixed in 3263c35

roles/splunk/tasks/kvstore_upgrade.yml
arcsector: Fixed in https://github.com/splunk/ansible-role-for-splunk/commit/3263c35509de7ece33be496ed9f6b0fbead62f94

roles/splunk/defaults/main.yml
```diff
@@ -69,6 +70,10 @@ splunk_shc_target_group: shc
 splunk_shc_deployer: "{{ groups['shdeployer'] | first }}" # If you manage multiple SHCs, configure the var value in group_vars
 splunk_shc_uri_list: "{% for h in groups[splunk_shc_target_group] %}https://{{ hostvars[h].ansible_fqdn }}:{{ splunkd_port }}{% if not loop.last %},{% endif %}{% endfor %}" # If you manage multiple SHCs, configure the var value in group_vars
 start_splunk_handler_fired: false # Do not change; used to prevent unnecessary splunk restarts
+splunk_enable_kvstore: true
+splunk_kvstore_storage: undefined # Can be defined here or at the group_vars level - accepted values: "wiredTiger" or "undefined", which leaves as default
+splunk_kvstore_version: undefined # Can be defined here or at the group_vars level - accepted values: 4.2 or "undefined", which leaves as default1
```
jewnix, line 75: I don't see this variable used either

arcsector: It's used here: https://github.com/splunk/ansible-role-for-splunk/pull/177/commits/9f689aa08ddf78dd87a56cde663ff497a42dc920#diff-0977dd8c4cb1bc6903196d2c0c0efd06d11183fbdb1dccf3620ffff234438804R28

arcsector: Oh I see, the `splunk_kvstore_version` is unused - I'll add it to the conditionals for the bottom of the upgrade procedure

jewnix:
After some preliminary testing, there are some issues that need to be addressed here.
1. I think the `adhoc_destructive_resync_kvstore.yml` task should be removed from the PR. Some things there do not completely make sense to me. Maybe I'm understanding it wrong, but I don't think you need to remove a member from the SHC to do that.
2. The `kvstore_upgrade.yml` does not work on version 8.2.x because the `splunk_kvstore_version` does not return anything, so it fails on the conditional for the block.
3. Leaving the `splunk_kvstore_storage` as undefined, also will not update the engine because of the conditionals on the block level. At least on the 8.2 that I tested.

arcsector:
@jewnix My thoughts:
1. Destructive resync is something that has been recommended to me through Splunk Support multiple times in order to fix KV issues, so I included it here - it seems that the advantage to it over just a normal clean is that it pulls both a fresh copy of the SH Bundle as well as the KV bundle, so that's the benefit there.
2. Good call, I'll put a default value in there for versions <= 8.2
3. That's what I want, I want people to have to specify they're upgrading in order to be able to do so one cluster component at a time.

dtwersky:
> * Destructive resync is something that has been recommended to me through Splunk Support multiple times in order to fix KV issues, so I included it here - it seems that the advantage to it over just a normal clean is that it pulls both a fresh copy of the SH Bundle as well as the KV bundle, so that's the benefit there.

So here is what I think. The destructive resync should be removed from this PR. 1. Because this is a snowflake issue, and destructive KVStore sync is not something that is documented. 2. Because this also destroys the SHC completely.
Even though support sometimes tells you sometimes to do it, does not mean it's something that should be done.

arcsector:
@dtwersky @jewnix Sorry I've been inactive on this, I removed the destructive resync, and I added a default value, though it's not for `splunk_kvstore_version`, rather for the `splunk_current_server_version_out.stdout` check, where the former is powered by default vars, and the latter is performed by a CLI call

arcsector:
Updating this with oplog size increase, as well as some helpful tasks to get KVStore-status and SHCluster-status as JSON blobs for ansible consumption.

dtwersky:
Hi @arcsector , 

Sorry this was left dormant for so long after so much work has gone in to this. I have been working internally to figure out all of this for a while on a different project, that was more for ephemeral docker instances, but I discovered a lot of things related to this PR that made me look at KVStore upgrades a little differently. There are so many differences between Splunk versions, MongoDB versions and MongoDB engines regarding to upgrade paths. I'm not sure if we should assume that people are still running version 8, and because later versions already automatically migrate and update, there may only be a need to run some of these commands in specific scenarios only.

There are so many amazing things in this PR, and I don't want to close this out and start from fresh, but maybe it needs to be revisited, and think if we want to make this compatible with older versions, or major version jumps.

What are your thoughts?

arcsector:
Thanks so much for the positive comments, glad you like the materials here - I'm definitely open to revisiting this as a PR of optional tasks and then making a playbook that calls all of them to do an all-in-one upgrade. Does that sound like a good plan - I could even put them in a sub-folder `roles/splunk/tasks/kvstore/...` if that would help to centralize this.

Do you happen to have a good map of those version transitions and what they entail as far as mongod version and engine? I'm having to go through the docs and switch back and forth between versions, as it's not clear what the approach should even be... Thanks!

PR #182, opened 2023-04-24
Extend configure license task

schneewe:
In this PR are some changes:
1. change yes boolean to true, because of yamllint rule
2. Adjust the step of remove the manager_uri setting when using local license. With a specific condition, e.g. upgrading to a 9.x version and want to use the newer "manager_uri" and changing to local license. It will never remove the old master_uri setting, because of the mode_option set to manager_uri. So I think it's better to have both variants in the removing step.
3. Include the check, if the general_key changed, so we have no "changed" task like in configure_deploymentclient.conf


schneewe:
@dtwersky 
Thanks for the hint, didn't look up before.
Added the README.md changes to this PR

PR #187, opened 2023-06-01
Change log level in cfg and CLI

arcsector:
# Summary

Changes the log-level at the CLI level, then makes sure `log-local.cfg` exists, and adds or modifies the the log level in there. Requires the variables `splunk_component` and `splunk_component_log_level` provided to the task.

## Modified files

- Added: `roles/tasks/adhoc_change_log_level.yml`
- Changed: `README.md`

### DIFF
--- a/roles/splunk/tasks/adhoc_change_log_level.yml
+++ b/roles/splunk/tasks/adhoc_change_log_level.yml
@@ -1,0 +1,44 @@
+---
+- name: Change log level of {{ splunk_component }} to {{ splunk_component_log_level }} via CLI
+  ansible.builtin.command: |
+    {{ splunk_home }}/bin/splunk set log-level {{ splunk_component }} -level {{ splunk_component_log_level }} -auth '{{ splunk_auth }}'
+  register: log_level_chg_out
+  changed_when: log_level_chg_out.rc == 0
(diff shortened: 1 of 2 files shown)


roles/splunk/tasks/adhoc_change_log_level.yml
```diff
@@ -0,0 +1,45 @@
...
+
+- name: Copy log to log-local if it doesn't exist
+  ansible.builtin.copy:
+    remote_src: true
+    src: "{{ splunk_log_cfg }}"
+    dest: "{{ splunk_log_local_cfg }}"
+    owner: "{{ splunk_nix_user }}"
+    group: "{{ splunk_nix_group }}"
+    mode: preserve
+  become: true
```
jewnix, line 33: missing `become_user`

jewnix:
@arcsector 
1. The `splunk set log-level` is not persistent, and is only used to change the level on a running instance temporarily.
2. Can you add the option to configure multiple components to different levels in a loop?

arcsector:
1. Yes, this is the desired functionality, change the level now, and then do it persistently so a restart isn't necessary. I can remove the restart handle to reflect this.
2. Done.

PR #191, opened 2023-06-07
Fixed the issue with main.yml

VatsalJagani:
A block of code that does not execute for Splunk full even though it is supposed to be executed.

### DIFF
--- a/roles/splunk/tasks/main.yml
+++ b/roles/splunk/tasks/main.yml
@@ -60,13 +60,6 @@
         systemd_unit: "{{ systemd_unit_uf }}"
       changed_when: false
 
-    - name: Setting least priviledged mode
-      set_fact:
-        least_privileged: true
-      when: >
-        (splunk_package_version is version(9.0, '>=') and splunk_install_type == "uf") or
-        splunk_install_type == "full" and splunk_nix_user == "root"
-
     - name: Configure var for splunkforwarder systemd service handler
       tags: always
       set_fact:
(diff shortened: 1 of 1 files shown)


jewnix:
Actually, the least-privileged setting only applies to the UF, and it does not work yet with a full installation anyway, and I am not sure if it's in the pipeline, that's why it's in the UF variables block.

Given all that, the `when` clause should probably be changed to 
```
when: (splunk_package_version is version(9.0, '>=') and splunk_nix_user != "root")
```
that will look a lot cleaner nad makes a lot more sense. But it needs to be tested to make sure nothing breaks.

VatsalJagani:
@jewnix - Why it does not apply to full installation?
- What if the user wishes to collect the /var/logs from Splunk SH and IDX instances as well?

jewnix:
@VatsalJagani `least-privileged` is not available on a full installation. This is a product issue.

On a full installation, we run the `configure_facl.yml` task to allow the splunk user to access `/var/log`. That task will only run when `not least_privileged`.

jewnix:
@VatsalJagani 

I'm not sure I understand the question, but `least_privileged` is set to `false` by default [here](https://github.com/splunk/ansible-role-for-splunk/blob/master/roles/splunk/defaults/main.yml#L16). This will automatically set a full installation or a UF running as root to false, and only a UF running version >= 9 that is running as a non-root user will get the `least_privileged = true` variable.

The line you are referring to in the `remove_splunk.yml` is there to specify to only remove those settings when splunk is not running as root or it is a full install.

VatsalJagani:
@jewnix  My question simplified: 
* Lets say if we remove the code block I changed, it does not make any change to execution for any tasks.

jewnix:
@VatsalJagani it does, it changes the value to true for UF v9, and the the `configure_facl.yml` task will skip instead of run. If it's running in least privilege mode, that task does not need to run.

PR #207, opened 2023-09-16
add auditd filtering tasks/vars

zyphermonkey:


roles/splunk/tasks/configure_auditd.yml
```diff
@@ -0,0 +1,45 @@
+---
+- name: "get {{ splunk_nix_user }} user uid"
+  getent:
+    database: passwd
+    key: "{{ splunk_nix_user }}"
+
+# - debug:
```
dtwersky, line 7: Please remove this and other commented out lines

roles/splunk/tasks/configure_auditd.yml
```diff
@@ -0,0 +1,45 @@
+---
+- name: "get {{ splunk_nix_user }} user uid"
```
dtwersky, line 2: Will this work if `splunk_nix_user` is not a local account?

roles/splunk/tasks/configure_auditd.yml
```diff
@@ -0,0 +1,45 @@
...
+  become: true
+  template:
+    src: 20-splunk.rules.j2
+    dest: /etc/audit/rules.d/20-splunk.rules
+    mode: 0600
+    owner: root
+    group: root
+  register: splunk_rule
+
+- name: Get auditd enabled level  # noqa 301 306
```
dtwersky, line 20: Do we need the `noqa`?

roles/splunk/tasks/configure_auditd.yml
```diff
@@ -0,0 +1,45 @@
...
+  become: true
+  shell: auditctl -s | grep enabled | cut -d" " -f2
+  changed_when: false
+  check_mode: false
+  register: auditctl_enabled
+
+# restart auditd if not immutable
+- name: restart auditd if not immutable  # noqa no-handler
```
dtwersky, line 28: when `splunk_rule is changed` `auditd` should be able to be reloaded without a restart, only enabling auditd needs a restart.
The restart and reload should probably be two separate tasks.

roles/splunk/tasks/configure_auditd.yml
```diff
@@ -0,0 +1,45 @@
+---
+- name: "get {{ splunk_nix_user }} user uid"
+  getent:
+    database: passwd
+    key: "{{ splunk_nix_user }}"
+
+# - debug:
+#     msg: "{{ splunk_nix_user }} user id = {{ ansible_facts.getent_passwd[splunk_nix_user][1] }}"
+
+- name: auditd - set 20-splunk.rules
```
dtwersky, line 10: Can we have the whole task only run if `auditd` is installed? If it's not installed, the task will fail.

roles/splunk/tasks/configure_auditd.yml
```diff
@@ -0,0 +1,45 @@
...
+    mode: 0600
+    owner: root
+    group: root
+  register: splunk_rule
+
+- name: Get auditd enabled level  # noqa 301 306
+  become: true
+  shell: auditctl -s | grep enabled | cut -d" " -f2
+  changed_when: false
+  check_mode: false
```
dtwersky, line 24: Why is `check_mode` set to false?

roles/splunk/tasks/configure_auditd.yml
zyphermonkey: I'll take care of this in the next commit.

roles/splunk/tasks/configure_auditd.yml
zyphermonkey: I never even thought about that. Is that even something we have to worry about. The role has always created local accounts and we don't use ldap atm for a central account repository. 

I'll reach out to some of our other teams I know that have non-local accounts and see what I can find out.

roles/splunk/tasks/configure_auditd.yml
zyphermonkey: The noqa is for ansible-lint. It disables specific checks on a task.  
https://ansible.readthedocs.io/projects/lint/usage/#muting-warnings-to-avoid-false-positives

Since this repo isn't using ansible-lint atm I'll get this removed to avoid any confusion.

roles/splunk/tasks/configure_auditd.yml
zyphermonkey: reloading auditd is only possible if it's not configured to be immutable. It can still be reloaded, but no changes are applied until reboot. Because of this we don't want ansible to report "changed" on a task when it never successfully applied anything giving the user a false sense of completion.

We also can't in good conscious be applying reboots to machines. This is the best approach we've found which is to apply if we can (auditd is not immutable) or notify user via debug msg that a reboot is required (auditd is immutable). 

We've had these tasks in place for a long time now and they're really solid.

roles/splunk/tasks/configure_auditd.yml
zyphermonkey: This task is just setting a variable, but ansible won't run `shell/command` tasks in check_mode. To ensure we don't get variable undefined on later tasks with `when` conditionals we have to force this task to run in `normal` and `check` modes.

>To force a task to run in normal mode and make changes to the system, even when the playbook is called with `--check`, set `check_mode: false`.   

https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_checkmode.html

roles/splunk/tasks/configure_auditd.yml
zyphermonkey: I think this is already solved for in `roles/splunk/defaults/main.yml` with the default being `splunk_auditd_configure: false`

This gives the user the ability to enable these tasks, but they aren't enabled by default. Hopefully a user won't enable them if they aren't running an OS that uses auditd.

dtwersky (approved):

PR #219, opened 2024-01-04
Add ambientcapability to service for dmesg

schneewe:
This change allows Splunk to read kernel log and use dmesg even with dmesg_restrict active. Therefore no need to activate for all non-root users


dtwersky:
@schneewe There is a configurable [var](https://github.com/splunk/ansible-role-for-splunk/blob/master/roles/splunk/defaults/main.yml#L78) to enable `dmesg` permissions using `sysctl` (default disabled), which will be needed if someone is using init.d.

Some people may not want to enable this feature for security purposes.

jewnix:
What I meant is that some users may not want splunk to have access as well.

PR #229, opened 2024-10-25
do basic linting (indent,fqcn,octals,truthy)

PerfectlyColouredMoose:
using yamllint (1.35.1) and ansible-lint (24.9.0, ruamel-yaml 0.18.6) to enhance the code standard:
* providing `.yamllint`, `.ansible-lint` files to set some chosen baseline
* fixed some bad/inconsistent indentations
* using everywhere the fully-qualified collection names (FQCN)
* declared and corrected truthyness to only allow true/false/True/False
* replaced usage of octal values by the recommended string versions
* optional .pre-commit file can be used for further changes


dtwersky:
@PerfectlyColouredMoose There is old a `.yamllint` [here](https://github.com/splunk/ansible-role-for-splunk/blob/master/roles/splunk/.yamllint), can you merge/remove the old one?

dtwersky:
@PerfectlyColouredMoose Can you please add some information and basic instructions to the README about linting and that hey should add a plugin to whatever IDE they are using, as well as the requirements and behavior pre-commit will have?
The main thing I'm concerned about is the pre-commit, which may throw off some users who are not used to it.

.yamllint.yml
```diff
@@ -0,0 +1,43 @@
...
+  line-length: disable
+  # When using `ansible-galaxy init ..` the newline is removed
+  new-line-at-end-of-file: disable
+  new-lines: enable
+  octal-values:
+    forbid-explicit-octal: true
+    forbid-implicit-octal: true
+  quoted-strings: disable
+  trailing-spaces: enable
+  truthy: { allowed-values: ["True", "False", "true", "false"] }
```
dtwersky, line 43: We should maybe stick to a single case. I would say to stick to lowercase

dtwersky, line 43:
```suggestion
  truthy: { allowed-values: ["true", "false"] }
```

dtwersky (approved):

PR #232, opened 2025-01-23
Configure Splunk Launch explicitly to prevent future clobbering

arcsector:
# Summary
This PR is needed because of the changes Splunk introduced in v9.1.0 of the UF, which [changed the default username of the Splunk UF installation](https://community.splunk.com/t5/Splunk-Enterprise/Why-did-Universal-forwarder-9-1-0-linux-change-owner/m-p/655673) to `splunkfwd` from `splunk`. This raised awareness of the fact that we can't rely on Splunk to keep its own internal values consistent, and as such explicitly specifying the user and service name will help prevent these values from being changed in the future.

This PR also adds in additional functionality for those who would like to store their indexes in a different location. A use case for this would be if you want to run your indexes on slower SSDs than your main application, you would set SPLUNK_HOME to be on your root FS and then have your `SPLUNK_DB` pointing to your other mounted SSDs like at `/data/splunk/db`. This was done as it seemed like the right place to put this type of feature.

### Added:
- Splunk launch config setting Splunk user, home, service name, and DB path if defined
- New var for DB Path

### Changed:
- README references new Splunk launch task
- installation task includes Splunk launch task


dtwersky:
@arcsector I'm pretty sure that the `splunk-launch.conf` will populate the user home and service field when enabling boot-start with the proper variables. Splunk is only using `splunkfwd` when installing using the `.deb` or `.rpm` file. In the tarball the `SPLUNK_OS_USER` is unset.
Are you trying to update a splunk installation that was originally installed using a package manager?

 [Here](https://github.com/splunk/ansible-role-for-splunk/blob/master/roles/splunk/tasks/configure_splunk_boot.yml#L56) is the command, and AFAIK that updates the file as well. 

The `SPLUNK_DB` is very useful though.

When updating Splunk with a different user, the same _should_ happen because we are disabling and re-enabling boot-start.

The only time I think this may cause issues is when using SysV, and for them, I don't care :laughing:

arcsector:
Yeah you're right about the splunkfwd user only being present in the packages not in the tarball. The problem is that in any production environment you're throwing the packages up to either a custom package repo or a ready-made repo like artifactory and that's how you're managing the versions, most people aren't managing these with tarballs, so that's the first reason I have to still include this.

Secondarily, the `splunkfwd` user isn't the only reason, it's just the most egregious example from Splunk - another reason might be, say, if Splunk does this silent user rollout again but with the tarball. Another would be if you want to go from running as root to running as a non-privileged user. As it stands, you'd have to uninstall splunk and reinstall, as I think [the boot start command only sets that value if it is currently unset](https://docs.splunk.com/Documentation/Splunk/9.4.2/Admin/Splunk-launchconf), but I could be wrong. Regardless, you shouldn't have to disable boot start and then reenable it in order to convert the Splunk user to a non-root user.

I am unsure if this is cause enough to include that, though. So if not, I'm okay with editing the PR to just keep the DB location and then we can remove the other functionality.

dtwersky:
I don't think we need to manually add things to that file. The `SPLUNK_HOME` is set on startup (or ftr?), so there is no need to add it there.
The `SPLUNK_SERVER_NAME` can be changed by setting the [unit file name](https://github.com/splunk/ansible-role-for-splunk/blob/master/roles/splunk/defaults/main.yml#L45-L46). All of these things are automatically updated in `splunk-launch.conf` when you start splunk or reconfigure Splunk, and having multiple places where they can be modified may just add confusion.

arcsector:
Done and this is also synced up to master as well.

PR #241, opened 2025-06-09
Rhel10 and SELinux compatibility

observable-it:
RHEL10:
I'm testing this role on a RHEL 10 VM. This role crashes on RHEL 10 because the syslog logrotation file is called differently: /etc/logrotate.d/rsyslog instead of /etc/logrotate.d/syslog. When testing, I discovered that the main.yml task contains the wrong order of "with_first_found", first it finds RedHat10.0, then RedHat, then RedHat10 then RedHat. As a result, only the RedHat.yml file is used, which contains the wrong "logrotate_file" variable.  

SELinux: 
Sometimes Splunk wouldn't start after installing or upgrading when SELinux is in enforcing mode (common issue on RHEL based systems).
On RHEL-based systems with SELinux enabled, extracting files from a tarball (e.g., using `tar -xvf`) does not automatically restore the correct SELinux security contexts for the extracted files. This can result in Splunk not starting or working correctly after install or upgrading. To resolve this, a `restorecon -R` task is added after extraction to apply the expected SELinux contexts based on policy. 

Requirements files.
RHEL 10 includes a reduced set of default Ansible collections and Python modules. Explicit requirements files are added to ensure all necessary dependencies are available.
Not sure if this is a desired way of maintaining dependencies for this repository (maybe people prefer adding requirements to the readme or environment)


observable-it:
@dtwersky Thanks for your feedback, I've removed the ansible requirements commit

dtwersky (approved):

PR #246, opened 2025-09-09
Fix regex pattern for extracting Splunk version

tomtrenz:


The5imon:
https://github.com/splunk/ansible-role-for-splunk/blob/784d2377a3f7b410c82283bf443d1bd54097b035/roles/splunk/tasks/check_splunk.yml#L49

Should also be adjusted to the following to make it more robust.

```yaml
when: splunk_version_release is version(splunk_package_version, '<')
```

This was also used in other version comparisons in the repo.

PR #252, opened 2026-01-27
testing: add Docker + Molecule local harness

kundeng:
## Why
This repo does not currently ship a repeatable local integration test harness for validating role changes against realistic, systemd-based Splunk topologies.
This PR adds an isolated `testing/` framework that spins up container topologies and runs Molecule scenarios in phases, keeping all scaffolding out of production role paths.

## What
- Add `testing/` with:
  - Taskfile workflows (`task infra:test`, `task day0:test`, `task day1:test`, `task workflow:full`)
  - Molecule scenarios: `testing/molecule/{infra,day0,day1}`
  - Environment inventories: `testing/molecule/environments/{small,dev,prod}`
  - Docker images for systemd+sshd hosts, ansible-controller, molecule-runner, optional git-server

## Web terminal (password-protected)
- `task infra:setup` exposes a local web terminal at `http://localhost:3000/ttyd`.
- Auth is enabled via ttyd `-c ansible:<password>`.
- The password is read from `testing/.secrets/ansible_password` (generated by `task setup:secrets`, not committed).

## Design notes
- Parameterized environments via `MOLECULE_ENV` to avoid scenario duplication while supporting multiple topologies.
- Infra prepare uses docker-connection as root + `ansible_remote_tmp=/tmp/.ansible/tmp` to avoid intermittent temp-dir failures in minimal/systemd containers.
- Infra verify checks PID1 via `/proc/1/comm` to avoid assuming `ps` exists in base images.

## Security / scope
- No production code paths changed; everything is under `testing/`.
- Test-only credentials exist in lab config/images and are not intended for production use.

## How to run
- `cd testing && task setup`
- `cd testing && task infra:test`
- `cd testing && task workflow:full`
- `cd testing && MOLECULE_ENV=prod task infra:test`

## Follow-ups (out of scope)
- Add a scenario that exercises git-based app deployment (`roles/splunk/tasks/configure_apps.yml`) against the test git-server.


dtwersky (commented):
please review the comments

testing/molecule/environments/small/group_vars/all.yml
```diff
@@ -0,0 +1,55 @@
...
+
+# Splunk configuration - adapted for testing
+splunk_home: /opt/splunk
+splunk_install_path: /opt
+splunk_nix_user: splunk
+splunk_nix_group: splunk
+splunk_nix_home: /home/splunk  # Use splunk user home
+splunk_admin_username: admin
+splunk_admin_password: "Password123!"  # Test password
+splunk_auth: "{{ splunk_admin_username }}:{{ splunk_admin_password }}"
```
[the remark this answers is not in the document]
kundeng: Fixed. `setup:secrets` now generates `testing/.secrets/splunk_admin_password` alongside the existing SSH and terminal secrets. The group vars now resolve it as:

```yaml
splunk_admin_password: "{{ lookup('env', 'SPLUNK_ADMIN_PASSWORD') | default(lookup('file', '/workspace/testing/.secrets/splunk_admin_password') | trim, true) }}"
```

Same pattern applied to `environments/dev/group_vars/all.yml` and `environments/prod/group_vars/all.yml`. The value comes from the environment variable when explicitly set (CI-friendly), and falls back to the generated file for local runs — no hardcoded literal in either path.

testing/molecule/infra/create.yml
```diff
@@ -0,0 +1,83 @@
+---
+# Create containers using Ansible
+# Dynamically creates containers based on environment inventory
+- hosts: localhost
+  gather_facts: false
+  vars:
+    targets: "{{ groups['full'] | default([]) + groups['git_server'] | default([]) + groups['ansible_controller'] | default([]) }}"
+  tasks:
+    - name: Set Python interpreter for Docker modules
+      set_fact:
```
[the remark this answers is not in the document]
kundeng: Fixed. The `set_fact` task is removed and `ansible_python_interpreter` is now defined at play scope instead:

```yaml
- hosts: localhost
  gather_facts: false
  vars:
    ansible_python_interpreter: /usr/local/bin/python3
```

Applied consistently in both `create.yml` and `destroy.yml`.

testing/docker-images/almalinux9-systemd-sshd/Dockerfile
```diff
@@ -0,0 +1,115 @@
...
+password   sufficient   pam_unix.so sha512 shadow\n\n\
+# Session section - minimal required\n\
+session    optional     pam_unix.so\n\
+session    optional     pam_systemd.so' > /etc/pam.d/system-auth
+
+# Remove /run/nologin if it exists to allow logins immediately
+RUN rm -f /run/nologin
+
+# Ensure SSH server is properly configured for key authentication
+RUN echo "PubkeyAuthentication yes" >> /etc/ssh/sshd_config && \
```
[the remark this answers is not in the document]
kundeng: Good suggestion. The inline `RUN echo` approach works for this test-only image but you're right that COPY would be cleaner and more maintainable. I'll leave that as a follow-up refactor to keep this PR focused — the current inline form is self-contained which is convenient for a local harness image, but switching to COPY files in the build context is the right direction if these Dockerfiles grow further.

testing/Taskfile.yml
```diff
@@ -0,0 +1,461 @@
...
+          chmod 644 .secrets/id_rsa.pub
+          echo "✅ SSH secrets created"
+        fi
+
+  setup:images:
+    desc: Build all Docker images required for testing (idempotent)
+    cmds:
+      - |
+        echo "🔧 Checking/building Splunk testing framework images..."
+        docker images | grep -q splunk-base-almalinux9 || docker build -t splunk-base-almalinux9:latest {{.DOCKER_IMAGES_DIR}}/almalinux9-systemd-sshd/
```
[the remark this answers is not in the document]
kundeng: Fixed. All grep checks now include `:latest` so they only match the exact tagged image and won't falsely match unrelated images with a similar name prefix:

```bash
docker images | grep -q splunk-base-almalinux9:latest || docker build ...
```

Applied to both the conditional build lines and the `status:` idempotency checks.

testing/Taskfile.yml
```diff
@@ -0,0 +1,461 @@
...
+      - docker images | grep -q splunk-ansible-controller
+      - docker images | grep -q splunk-git-server
+      - docker images | grep -q molecule-runner
+
+  download:splunk:
+    desc: "Download Splunk binaries to testing/software directory (idempotent)"
+    dir: "{{.PROJECT_ROOT}}/testing"
+    cmds:
+      - |
+        if [ -f software/splunk-9.1.2-b6b9c8185839-Linux-x86_64.tgz ] && [ -f software/splunkforwarder-9.1.2-b6b9c8185839-Linux-x86_64.tgz ]; then
```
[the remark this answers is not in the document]
kundeng: Fixed. Version and build are now driven by `SPLUNK_VERSION` and `SPLUNK_BUILD` environment variables with the current test defaults preserved as fallbacks:

```bash
SPLUNK_VERSION="${SPLUNK_VERSION:-9.1.2}"
SPLUNK_BUILD="${SPLUNK_BUILD:-b6b9c8185839}"
```

The same env vars are passed through to every `docker run` call for the Molecule scenarios, and the group vars derive `splunk_package_version`, `build_id`, and `splunk_file` from them. Changing the version for a test run is now a single env var override with no file edits needed.

testing/docker-images/ansible-controller/Dockerfile
```diff
@@ -0,0 +1,62 @@
...
+FROM ubuntu:22.04 AS builder
+
+# Install build dependencies
+RUN apt-get update && \
+    DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends \
+      build-essential cmake git ca-certificates libjson-c-dev libwebsockets-dev && \
+    apt-get clean && rm -rf /var/lib/apt/lists/*
+
+# Clone and build ttyd from source (latest version 1.7.7)
+RUN git clone https://github.com/tsl0922/ttyd.git /tmp/ttyd && \
```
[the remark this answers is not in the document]
kundeng: Fixed. Added an explicit `ARG TTYD_VERSION=1.7.7` and the clone now targets that tag with `--depth 1` for a reproducible build:

```dockerfile
ARG TTYD_VERSION=1.7.7
RUN git clone --branch ${TTYD_VERSION} --depth 1 https://github.com/tsl0922/ttyd.git /tmp/ttyd
```

On the question of why we build from source rather than using the OS package: the Ubuntu 22.04 package for ttyd is significantly older (1.6.x), and the version in the OS repos doesn't support the `--writable` flag or the theme options we use for the web terminal. Building from source gives us the current stable release pinned explicitly.

testing/molecule/environments/prod/group_vars/all.yml
```diff
@@ -0,0 +1,67 @@
...
+splunk_shc_rep_port: 8100
+splunk_shc_target_group: shcluster
+
+# Test versions - using stable versions for testing
+splunk_version: "{{ splunk_test_version }}"
+# Local software paths (predownloaded to testing/software/)
+splunk_package_path_full: "/workspace/testing/software/splunk-{{ splunk_test_version }}-{{ splunk_test_build }}-Linux-x86_64.tgz"
+splunk_package_path_uf: "/workspace/testing/software/splunkforwarder-{{ splunk_test_version }}-{{ splunk_test_build }}-Linux-x86_64.tgz"
+# Remote URLs as fallback
+splunk_package_url_full: "https://download.splunk.com/products/splunk/releases/{{ splunk_test_version }}/linux/splunk-{{ splunk_test_version }}-{{ splunk_test_build }}-Linux-x86_64.tgz"
```
dtwersky: These urls will not work on Splunk version > 9.3. The URL was changed to `*linux-amd64.tgz`

See [main.yml](https://github.com/splunk/ansible-role-for-splunk/blob/master/roles/splunk/defaults/main.yml#L18-L20)

testing/molecule/day0/converge.yml
```diff
@@ -0,0 +1,77 @@
+---
+# Molecule converge playbook for day0
+# Installs Splunk on target hosts using local software
+
+# Copy Splunk software to target hosts
+- name: Copy Splunk software to target hosts
+  hosts: all:!ansible_controller:!git_server:!jumpbox
+  vars:
+    splunk_test_version: "{{ lookup('env', 'SPLUNK_VERSION') | default('9.1.2', true) }}"
```
dtwersky: version 9.1 is EOL, make the default the latest version, You can probbably also use the default in default/main.yml, so you dont have to specify it every time, That has the default (I should update that as well to the latest)

testing/molecule/day0/converge.yml
```diff
@@ -0,0 +1,77 @@
...
+    splunk_test_build: "{{ lookup('env', 'SPLUNK_BUILD') | default('b6b9c8185839', true) }}"
+  tasks:
```
dtwersky: Please use FQCN for all Ansible modules (`ansible.builtin.*`) throughout the `testing/molecule/` directory. Current violations:

- `testing/molecule/infra/prepare.yml` - 36 occurrences (e.g. `file:`, `copy:`, `systemd:`, `user:`, `lineinfile:`, `slurp:`, `set_fact:`, `stat:`, `assert:`, `wait_for:`, `blockinfile:`)
- `testing/molecule/day1/verify.yml` - 9 occurrences (`stat:`, `set_fact:`, `debug:`, `assert:`)
- `testing/molecule/day0/verify.yml` - 8 occurrences (`systemd:`, `wait_for:`, `uri:`, `debug:`)
- `testing/molecule/infra/converge.yml` - 5 occurrences (`wait_for:`, `systemd:`, `file:`, `copy:`, `blockinfile:`)
- `testing/molecule/infra/verify.yml` - 2 occurrences (`debug:`, `assert:`)

Also, please run `ansible-lint` on the `testing/` directory and fix any issues before merging.

testing/docker-images/almalinux9-systemd-sshd/.dockerignore
```diff
@@ -0,0 +1,4 @@
+.git
+.gitignore
+README.md
+*.md
```
dtwersky: no EOF

testing/CLAUDE.md
```diff
@@ -0,0 +1,371 @@
+# CLAUDE.md
```
dtwersky: Please remove this file

testing/PROGRESS.md
```diff
@@ -0,0 +1,115 @@
+# PROGRESS.md - AI Progress Tracking
```
dtwersky: Please remove this file

testing/docker-images/molecule-runner/Dockerfile
```diff
@@ -0,0 +1,44 @@
...
+RUN ansible-galaxy collection install ansible.posix && \
+    ansible-galaxy collection install community.crypto && \
+    ansible-galaxy collection install community.general && \
+    ansible-galaxy collection install community.docker
+
+# Create workspace
+WORKDIR /workspace
+
+# Default command
+CMD ["molecule", "--help"]
```
dtwersky: no EOF

testing/docker-images/ubuntu2204-systemd-sshd/.dockerignore
```diff
@@ -0,0 +1,4 @@
+.git
+.gitignore
+README.md
+*.md
```
dtwersky: no EOF

testing/molecule/environments/prod/group_vars/alma_hosts.yml
```diff
@@ -0,0 +1,3 @@
+---
+# AlmaLinux-specific configuration
+docker_image: splunk-base-almalinux9:latest
```
dtwersky: no EOF

testing/.env.example
```diff
@@ -0,0 +1,6 @@
+# Twingate Configuration (Optional)
+# Uncomment and set your Twingate network and tokens for external access
+# If not set, Twingate connector will not be started
+# TWINGATE_NETWORK=your-network-name
+# TWINGATE_ACCESS_TOKEN=your-access-token
+# TWINGATE_REFRESH_TOKEN=your-refresh-token
```
dtwersky: no EOF

testing/Taskfile.yml
```diff
@@ -0,0 +1,505 @@
...
+
+  setup:
+    desc: Complete environment setup (secrets + images + software + network + runner)
+    cmds:
+      - task: setup:secrets
+      - task: setup:images
+      - task: download:splunk
+      - task: network:create
+      - task: runner:start
+      - echo "✅ Environment setup complete!"
```
dtwersky: Please remove all emojis

dtwersky (commented):
Please resolve all open issues. You can reach out to me on slack `@jewnix` if you have any questions

dtwersky (commented):
Ran the full test suite (`infra:test` -> `day0:test` -> `day1:test`) for all three environments. Results:

| Environment | Nodes | infra | day0 | day1 |
|-------------|-------|-------|------|------|
| **small** | 2 | PASS | PASS | PASS |
| **dev** | 1 | PASS | PASS | PASS |
| **prod** | 9 | PASS | FAIL | FAIL |

See inline comments for the two prod failures.

testing/molecule/environments/prod/hosts.yml
```diff
@@ -0,0 +1,100 @@
...
+                  name: test_search_app
+                - git_project: test/splunk_apps/
+                  git_version: main
+                  name: test_dashboard_app
+                - git_project: test/splunk_apps/
+                  git_version: main
+                  name: test_lookup_app
+              target_shc_group_name: shcluster
+
+        uf:
```
dtwersky: Bug: `uf` is nested under `full`, making `splunk-uf01` a member of both groups. The upstream role expects `uf` and `full` as sibling groups under `all:children:`. This causes day0 to fail on `splunk-uf01`:

```
TASK [Check if Splunk Enterprise is installed]
fatal: [splunk-uf01]: FAILED! => {"changed": false, "msg": "Permission denied"}
```

`uf:` should be dedented to the same level as `full:`:

```yaml
all:
  children:
    full:
      children:
        clustermanager:
        ...
    uf:
      hosts:
        splunk-uf01:
```

testing/molecule/environments/prod/group_vars/all.yml
```diff
@@ -0,0 +1,67 @@
+---
+# Main inventory variables - clean and reusable
+# SSH connection overrides are in scenario-specific group_vars
```
dtwersky: Bug: This file is missing SSH connection variables that `small` and `dev` environments both define. Without them, the day1 scenario (and any scenario without a layered inventory overlay) fails with SSH auth errors on all 9 hosts:

```
fatal: [splunkapp-prod01]: UNREACHABLE! => {"msg": "ansible@splunkapp-prod01: Permission denied (publickey,password)."}
fatal: [splunk-master]: UNREACHABLE! => {"msg": "ansible@splunk-master: Permission denied (publickey,password)."}
... (all 9 hosts)
```

The day0 scenario masks this because `day0/molecule.yml` layers `day0/inventory/group_vars/all.yml` (which has the SSH config) on top. day1 only uses the environment inventory.

Add the following (matching `small` and `dev` environments):

```yaml
# Connection method override - SSH for realistic testing
ansible_connection: ssh
ansible_host: "{{ inventory_hostname }}"
ansible_ssh_private_key_file: /workspace/testing/.secrets/id_rsa
ansible_ssh_common_args: '-o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null'
ansible_python_interpreter: /usr/bin/python3
```

dtwersky:
Suggestion: consider adding a `testing/run-tests.sh` wrapper script so contributors can run the full test suite without installing `task` (or anything else) on their workstation. Only Docker is needed.

I tested this against all three environments and it works (the script is how I found the two bugs above).

```bash
#!/usr/bin/env bash
# Run the testing framework inside a container. Requires only Docker on the host.
#
# Usage:
#   ./run-tests.sh              # Interactive shell with task available
#   ./run-tests.sh setup        # Run a specific task command
#   ./run-tests.sh infra:test   # Run infra test scenario
#   ./run-tests.sh --list       # List all available tasks
#
# Environment variables:
#   MOLECULE_ENV=prod ./run-tests.sh infra:test   # Override environment
set -euo pipefail

PROJECT_ROOT="\$(cd "\$(dirname "\$0")/.." && pwd)"
TESTING_DIR="\$PROJECT_ROOT/testing"

# Build molecule-runner image if needed
if ! docker image inspect molecule-runner:latest >/dev/null 2>&1; then
    echo "Building molecule-runner image..."
    docker build -t molecule-runner:latest "\$TESTING_DIR/docker-images/molecule-runner/"
fi

# Create network if needed
docker network inspect splunk-test-network >/dev/null 2>&1 || \
    docker network create splunk-test-network

# Install go-task inside the container at startup
INIT='sh -c "\$(curl -fsSL https://taskfile.dev/install.sh)" -- -d -b /usr/local/bin >/dev/null 2>&1'

# Mount the project at the SAME host path so that nested docker run commands
# (executed by the host Docker daemon via the mounted socket) resolve volume
# mount paths correctly.
DOCKER_ARGS=(--rm --network splunk-test-network)

# Allocate a TTY only when stdin is a terminal
if [ -t 0 ]; then
    DOCKER_ARGS+=(-it)
fi

DOCKER_ARGS+=(
    -v /var/run/docker.sock:/var/run/docker.sock
    -v "\$PROJECT_ROOT":"\$PROJECT_ROOT"
    -w "\$TESTING_DIR"
)

# Forward MOLECULE_ENV if set
if [ -n "\${MOLECULE_ENV:-}" ]; then
    DOCKER_ARGS+=(-e "MOLECULE_ENV=\$MOLECULE_ENV")
fi

if [ \$# -gt 0 ]; then
    exec docker run "\${DOCKER_ARGS[@]}" \
        molecule-runner:latest \
        bash -c "\${INIT} && exec task \\"\\\$@\\"" _ "\$@"
else
    exec docker run "\${DOCKER_ARGS[@]}" \
        molecule-runner:latest \
        bash -c "\${INIT} && echo 'task ready - run: task --list' && exec bash"
fi
```

Key design choice: the project is mounted at the **same host path** inside the container (not at `/workspace`). This is necessary because the Taskfile's nested `docker run` commands use `{{.PROJECT_ROOT}}` for volume mounts, and the host Docker daemon resolves those paths relative to the host filesystem.

jewnix:
@kundeng 

I made a little bit of changes to the MR. Here is what and why:

  ### Secrets Management - Redesigned
  **Was**: Individual plain text files read via `lookup('env')` with file fallback, passed as `-e` env vars on every docker run.
  **Now**: Single YAML vars file at `.secrets/inventory/group_vars/all.yml`, loaded as an Ansible inventory source via `--inventory` in molecule.yml.
  **Why**: The env var approach required every docker run to forward secrets. Loading as inventory group_vars makes them regular Ansible variables with zero plumbing.

  ### Secrets Created on the Host
  **Was**: `setup:secrets` ran inside a Docker container as root, creating root-owned files.
  **Now**: `run-tests.sh` creates secrets on the host before entering Docker.
  **Why**: Root-owned files require `sudo rm` to clean up.

  ### Version Management Simplified
  **Was**: `splunk_test_version`/`splunk_test_build` env var lookups aliased to `splunk_package_version`/`build_id`. Four variables for two values, defaulting to EOL version 9.1.2.
  **Now**: Direct `splunk_package_version: 9.4.2` and `build_id: e9664af3d956` in group_vars.
  **Why**: The indirection served no purpose. The role's defaults handle arch detection (`Linux-x86_64` vs `linux-amd64`) automatically.

  ### Download Playbook Replaces Bash Download
  **Was**: Taskfile bash script with hardcoded `Linux-x86_64` arch in URLs.
  **Now**: `molecule/download.yml` playbook that loads role defaults for correct URL/arch.
  **Why**: The role already has arch detection logic. Duplicating it in bash guaranteed divergence when Splunk changed URL format in 9.4.

  ### Prod hosts.yml - uf Group Fix
  **Was**: `uf` nested under `full`, making UF hosts members of both groups.
  **Now**: `uf` is a sibling of `full` under `all:children`.
  **Why**: The role expects separate groups. Nesting caused permission errors on `splunk-uf01`.

  ### Prod SSH Vars Added
  **Was**: Missing SSH connection variables. Day1 failed with auth errors on all 9 hosts.
  **Now**: SSH vars added to prod group_vars, matching small/dev.
  **Why**: Day0 masked the issue via scenario-specific inventory overlay. Day1 only loaded the environment inventory.

  ### prepare.yml - Secrets Read from Localhost
  **Was**: SSH keys read via `run_once` on a random target container. `/workspace` only exists on the molecule-runner. Failed ~90% of the time on prod.
  **Now**: Separate localhost pre-play reads secrets from the molecule-runner, exposes them via `hostvars['localhost']`.
  **Why**: `run_once` picks an arbitrary host. Reading on localhost is deterministic.

  ### veth Checksum Offloading Fix
  **Was**: 1.2GB tarball transfers via SSH corrupted intermittently (CRC errors).
  **Now**: `ethtool -K eth0 tx off rx off` runs inside every container. Containers use `--cap-add NET_ADMIN`.
  **Why**: Docker veth pairs have TX/RX checksum offloading enabled by default. The kernel marks packets as "hardware will compute checksum" but veth has no hardware - both ends skip validation, allowing silent data corruption. This is a [documented bug](https://github.com/kubernetes/kubernetes/issues/18898). It's old, but this fixed it. Go figure.

  ### Taskfile Refactored
  **Was**: 11 copy-pasted docker run blocks (448 lines), 78 lines of backward-compatible aliases.
  **Now**: `_DOCKER_ARGS`, `_MOL`, `_MOL_NET` variables (225 lines). Each task is a one-liner.
  **Why**: Adding a single env var required 10 edits. Aliases for `small` were unnecessary since it's already the default.

  ### Day1 Converge - Uses the Role Properly
  **Was**: Manually detected `splunk_service` via stat checks, bypassed `tasks/main.yml`. Didn't handle UF service names.
  **Now**: Uses the role via `deployment_task` variable, same pattern as day0.
  **Why**: The role's `tasks/main.yml` sets `splunk_service` correctly. Bypassing it meant duplicating and incorrectly reimplementing that logic.

  ### Day1 Verify - Independent of the Role
  **Was**: 85 lines with `ps aux | grep splunk` (fails on AlmaLinux without procps), only checked `Splunkd.service`.
  **Now**: 30 lines checking both `Splunkd.service` and `SplunkForwarder.service` via systemctl.
  **Why**: Verify must independently confirm system state. Using the role in verify is circular - `check_splunk_status.yml` has `failed_when: false` and never fails.

  ### AlmaLinux Dockerfile - COPY Instead of Inline Config
  **Was**: SSH/PAM config scattered across 7 RUN commands with echo/sed, duplicated (115 lines).
  **Now**: Config files in `config/` build context, 4 COPY commands (49 lines).
  **Why**: Reviewer requested it. Duplication caused confusion about which settings were in effect.

  ### Docker Volume Naming
  **Was**: Generic names (`gitea-data`) conflicting with user's existing containers. `docker volume prune -f` in reset.
  **Now**: Prefixed with `ansible-splunk-test-`. Reset only removes these specific volumes.
  **Why**: Generic prune destroys unrelated user data.

  ### go-task Baked into Image
  **Was**: `curl | sh` installed go-task on every container startup with no checksum or version pin.
  **Now**: Installed in molecule-runner Dockerfile at build time.
  **Why**: Supply chain risk and unnecessary latency on every run.

  ### Other Cleanup
  - Removed `testing/CLAUDE.md` and `testing/PROGRESS.md`
  - Fixed trailing newlines in 5 files
  - Removed all emojis from Taskfile
  - Added FQCN (`ansible.builtin.*`) to all molecule playbooks
  - Fixed `enabled: yes` -> `true` (yamllint truthy rule)
  - Fixed `with_items` -> `loop` (deprecated)
  - Removed dead `infra/converge.yml` (prepare.yml handles it)
  - `splunk_auth` in verify uses variable instead of hardcoded `admin:changeme`
  - Molecule-runner Dockerfile consolidated from 5 pip layers to 1, removed uv
  - Downloaded files chowned to host user UID/GID

  ### Test Results

  | Environment | Nodes | Serial | infra | day0 | day1 |
  |-------------|-------|--------|-------|------|------|
  | **small** | 2 | 50 | PASS | PASS | PASS |
  | **dev** | 1 | 50 | PASS | PASS | PASS |
  | **prod** | 9 | 50 | PASS | PASS | PASS |

dtwersky (approved):

PR #261, opened 2026-04-02
Update deploy app task and restart handler

mpham-splunk:
This PR creates an option in the restart Splunk handler in cases where you don't want Splunk to restart yet. In addition, I added retries option to the git clone task in the app deployment task as there can be transient ssh errors when performing the clone. I also added a task to set the ownership of the clone path so it's easier to synchronize the apps when running the task on the Ansible controller.

PR #264, opened 2026-04-13
Fix #263: move app ownership chown to target host, revert SHC check

### LINKED ISSUE #263: configure_apps stopped working after permissions set locally rearrangement
I am adopting the role some time ago and I use it mostly to deploy TAs to our infrastructure, with the help of AWX.

After PR #261  my setup stopped working.

From what I can see, the configure_apps now tries to set file permissions right after downloading git repository contents on this task "Set ownership of cloned repos to splunk user" and the task is run localy.

The problem here is that the host running ansible will only have minimum configurations and all ansible relative, not having splunk user or group, failing to run this step.

dtwersky:
## Summary

Fixes #263. The controller-side chown added in PR #261 fails on Ansible
controllers (e.g. AWX) that don't have a local splunk user. This PR moves
ownership enforcement back to the target host and addresses adjacent
regressions from PRs #261/#262.

## Changes

- **`configure_apps.yml`**
  - Remove the failing controller-side chown (fixes #263)
  - Remove the `'shdeployer' in group_names` SHC check block (moved back)
  - Add a post-loop remote chown that runs once per unique
    `splunk_app_deploy_path`, preserving the "not per-repo" optimization

- **`install_apps.yml`**
  - Restore `--no-owner` / `--no-group` in rsync_opts
  - Revert PR #262's SHC check move - block is back in the loop with the
    original per-app gate `when: handler == "apply shcluster-bundle"`.
    PR #262's `'shdeployer' in group_names` proxy was not semantically
    equivalent: it over-fired for shdeployers pushing only to `etc/apps`
    and silently skipped for non-shdeployers with a per-app override to
    `etc/shcluster/apps` (apps synced but SHC bundle never applied)
  - Add `| bool` on `service_ready_state` to fix the pre-existing
    string-truthiness bug

- **`handlers/main.yml`**: wrap `skip_splunk_restart` in `| bool`
- **`defaults/main.yml`**: declare `skip_splunk_restart: false` (PR #261
  introduced the toggle but never added it to defaults)

## Cost of the SHC revert

SHC plays now make N REST calls to the captain instead of 1. This matches
pre-PR-262 behavior that ran in production for years. Can be addressed in
a follow-up PR by gating on a precomputed set of effective deploy paths,
preserving the once-per-host optimization without the proxy's edge cases.

### DIFF
--- a/roles/splunk/defaults/main.yml
+++ b/roles/splunk/defaults/main.yml
@@ -83,6 +83,7 @@ splunk_shc_target_group: shc
 splunk_shc_deployer: "{{ groups['shdeployer'] | first }}"  # If you manage multiple SHCs, configure the var value in group_vars
 splunk_shc_uri_list: "{% for h in groups[splunk_shc_target_group] %}https://{{ hostvars[h].splunk_mgmt_uri }}:{{ splunkd_port }}{% if not loop.last %},{% endif %}{% endfor %}"  # If you manage multiple SHCs, configure the var value in group_vars
 start_splunk_handler_fired: false  # Do not change; used to prevent unnecessary splunk restarts
+skip_splunk_restart: false  # Set to true to suppress the "restart splunk" handler (e.g. during maintenance windows where you want to defer restarts)
 # Linux and scripting related vars
 add_crashlog_script: false  # Set to true to install a script and cron job to automatically cleanup splunk crash logs older than 7 days
 add_diag_script: false  # Set to true to install a script and cron job to automatically cleanup splunk diag files older than 30 days
--- a/roles/splunk/handlers/main.yml
+++ b/roles/splunk/handlers/main.yml
@@ -72,7 +72,7 @@
   become: true
   when:
     - not start_splunk_handler_fired
-    - not skip_splunk_restart | default(false)
+    - not (skip_splunk_restart | default(false) | bool)
 
 - name: restart redhat auditd service
   ansible.builtin.shell: |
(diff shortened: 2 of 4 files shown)


kdurril-splunk:
1. Chown is broad (`configure_apps.yml`, post-install chown): we can narrow the `path` to `{{ splunk_home }}/{{ item }}/{{ app_name }}` per unique app

or 

accept the broader chown AND ALSO add `changed_when: false` and document the behavior. The current combination of broad scope + no `changed_when: false` means every play run reports false positives as changed.

2. add `changed_when: false` on chown task: currently, the `file` task with `recurse: true` will report `changed` on every run since it re-evaluates all file attributes. Adding `changed_when: false` should suppress noise

dtwersky:
> 1. Chown is broad (`configure_apps.yml`, post-install chown): we can narrow the `path` to `{{ splunk_home }}/{{ item }}/{{ app_name }}` per unique app
> 
> or
> 
> accept the broader chown AND ALSO add `changed_when: false` and document the behavior. The current combination of broad scope + no `changed_when: false` means every play run reports false positives as changed.

I am aware of that. We accept that because splunk itself always tries to `chown splunk:splunk $SPLUNK_HOME`, so this is expected and accptable

mpham-splunk (approved):

dtwersky:
> 2. add `changed_when: false` on chown task: currently, the `file` task with `recurse: true` will report `changed` on every run since it re-evaluates all file attributes. Adding `changed_when: false` should suppress noise

I verified this empirically against a deploymentserver host - both the first deploy and subsequent reruns reported `changed: false` on this task (rsync writes as splunk_nix_user via become_user, so ownership is already correct and the recursive file check is a no-op). 

We also deliberately avoided `changed_when: false` here because the prior review flagged it as suppressing legitimate drift-detection signal - if the `chown` ever does report `changed: true`, that's real drift worth surfacing.

